PR #28, opened 2021-08-28
Add conversion tx for CRO tokens

thomas-nguy:
Add two msg in Cronos

- `MsgConvertVouchers` (address, coins) : 
         If the coin is the `IbcCroDenom` then it moves the amount from sender account to the module account, it mints 10^10 * amount `EvmDenom` to the module account and send them to the sender
         if the coin is in the list of `convertAllowedCoin`, it sends the token to smart contract account and mint ERC20 tokens (TODO)
         
- `MsgTransferTokens`  (from, to, coins)
        if the coin is `EvmDenom`, it burns the floor (amount div 10^10) from sender account and `transfer` the same amount of `IbcCroDenom` through IBC to the destination from module account
        if the coin is in the list of 'convertAllowedCoin' , it `transfer` the ibc tokens to Crypto.org  (only from smart contract account ) (TODO)
        
        
This PR focus only on the flow for CRO token conversion. I will open another PR to support IOU tokens from solo machine
          
TODO:

- [ ] fix tests
- [x] find correct client parameters to setup timeout height in IBC transfer
- [ ] add extra checks on from and to address when sending to crypto.org
- [x] test with solo machine

# PR Checklist:

- [] Have you included tests for any non-trivial functionality?
- [x] Have you checked your code formatting is correct? (`go fmt`)
- [x] Have you checked your basic code style is fine? (`golangci-lint run`)


x/cronos/keeper/msg_server.go
```diff
@@ -15,3 +22,139 @@ func NewMsgServerImpl(keeper Keeper) types.MsgServer {
...
+
+			// Send ibc token escrow address
+			err = k.bankKeeper.SendCoinsFromAccountToModule(ctx, acc, types.ModuleName, sdk.NewCoins(c))
+			if err != nil {
+				return nil, err
+			}
+			// Compute new amount, because basecro is a 8 decimals token, we need to multiply by 10^10 to make it
+			// a 18 decimals token
+			ten := big.NewInt(10)
+			exponent := ten.Exp(ten, ten, nil)
```
yihuang, line 52: How about make it a static variable?

x/cronos/keeper/msg_server.go
```diff
@@ -15,3 +22,139 @@ func NewMsgServerImpl(keeper Keeper) types.MsgServer {
...
+				ctx, types.ModuleName, acc, sdk.NewCoins(newAmount),
+			); err != nil {
+				return nil, err
+			}
+
+		default:
+			if err := k.IsConvertEnabledCoins(ctx, msg.Amount...); err != nil {
+				return nil, err
+			}
+			// TODO wrap to erc20 tokens
```
yihuang, line 73: Maybe we can do this in different PR, to keep this one simpler?

thomas-nguy: okay, I just finished this implementation but I added some todo

x/cronos/keeper/msg_server.go
thomas-nguy: thanks, I culdnt make as a constant but var works, fixed

proto/cronos/tx.proto
```diff
@@ -3,11 +3,38 @@ package cryptoorgchain.cronos.cronos;
...
+
 option go_package = "github.com/crypto-org-chain/cronos/x/cronos/types";
 
-// Msg defines the Msg service.
+// Msg defines the Cronos Msg service
 service Msg {
     // this line is used by starport scaffolding # proto/tx/rpc
+
+    // Send defines a method for converting ibc coins to Cronos coins.
+    rpc ConvertTokens(MsgConvertTokens) returns (MsgConvertResponse);
```
yihuang, line 16: I suppose we also need a message to convert erc20 back to native token and send through ibc channel?

proto/cronos/tx.proto
```diff
@@ -3,11 +3,38 @@ package cryptoorgchain.cronos.cronos;
...
+
+    // Send defines a method to send coins to Crypto.org chain
+    rpc SendToCryptoOrg(MsgSendToCryptoOrg) returns (MsgConvertResponse);
```
yihuang, line 19: Is this message only for sending CRO back to crypto-org, or sending any ibc tokens back?

proto/cronos/tx.proto
thomas-nguy: I think we can use "SendToCryptoOrg" message from smart contract 

For erc20: 
- ConvertToTokens will send ibc tokens to smart contract account and trigger mint erc20 in smart contract
- SendToCryptoOrg is trigger within smart contract is sending back ibc tokens from smart contract to user address and trigger IBC transfer

similarly to ethermint hook design

proto/cronos/tx.proto
thomas-nguy: both (that is why there is a switch in the logic) 
but we could separate them to different transaction if its too complex

proto/cronos/tx.proto
yihuang: I guess `SendToCryptoOrg` support any ibc channels, not limited to crypto-org then?
And is the `SendToCryptoOrg` triggered by smart contract using the evm hook proposal?

proto/cronos/tx.proto
yihuang: I see, maybe change the name to sth like `SendToIBC`, since we are not limited to crypto-org chain?

thomas-nguy: I though all IBC tokens from cronos will comes from crypto-org chain?

If its not the case, design might be even more complex...

Let just focus on the IBC CRO flow in this PR.

I think for the IBC <-> erc20, there are still few thinking that we need to clear

x/cronos/types/messages.go
```diff
@@ -0,0 +1,115 @@
...
+	return RouterKey
+}
+
+// Type ...
+func (msg MsgSendToCryptoOrg) Type() string {
+	return TypeMsgSendToCryptoOrg
+}
+
+// GetSigners ...
+func (msg *MsgSendToCryptoOrg) GetSigners() []sdk.AccAddress {
```
JayT106, line 84: Don't understand why return array?

x/cronos/types/messages.go
```diff
@@ -0,0 +1,115 @@
...
+	return RouterKey
+}
+
+// Type ...
+func (msg MsgConvertTokens) Type() string {
+	return TypeMsgConvertTokens
+}
+
+// GetSigners ...
+func (msg *MsgConvertTokens) GetSigners() []sdk.AccAddress {
```
JayT106, line 33: Don't understand why return array? The message doesn't use repeated param.

x/cronos/types/messages.go
thomas-nguy: The array comes from the interface
https://github.com/cosmos/cosmos-sdk/blob/master/types/tx_msg.go#L21

I think cosmos uses array to support multi sig?

proto/cronos/cronos.proto
```diff
@@ -0,0 +1,23 @@
...
+    string                  ibc_cro_channelID       = 3 [(gogoproto.moretags) = "yaml:\"ibc_cro_channelID,omitempty\""];
+}
+
+// ConvertEnabled maps coin denom to a convert_enabled status (whether a denom is
+// convertable).
+message ConvertEnabled {
+    option (gogoproto.equal)            = true;
+    option (gogoproto.goproto_stringer) = false;
+    string denom                        = 1;
+    bool   enabled                      = 2;
```
yihuang, line 22: Is this field necessary? can we remove the denom from the list when disabling?

proto/cronos/cronos.proto
```diff
@@ -0,0 +1,23 @@
...
+import "gogoproto/gogo.proto";
+
+option go_package = "github.com/crypto-org-chain/cronos/x/cronos/types";
+
+// Params defines the parameters for the cronos module.
+message Params {
+    option (gogoproto.goproto_stringer)             = false;
+    repeated ConvertEnabled convert_enabled         = 1 [(gogoproto.moretags) = "yaml:\"convert_enabled,omitempty\""];
+    string                  ibc_cro_denom           = 2 [(gogoproto.moretags) = "yaml:\"ibc_cro_denom,omitempty\""];
+    string                  ibc_cro_channelID       = 3 [(gogoproto.moretags) = "yaml:\"ibc_cro_channelID,omitempty\""];
```
yihuang, line 13: Is this duplicated, I think we can query channel id from the ibc token denom?

proto/cronos/cronos.proto
thomas-nguy: yes, in the future, we would have the mapping between denom <-> config (smart contract address, ibc config etc)

enable = false means that it is supported but not enable at the moment

proto/cronos/cronos.proto
thomas-nguy: good point, I will remove it

thomas-nguy:
Tested with solo machine, seems to work fine

proto/cronos/tx.proto
yihuang: User can create ibc channels to whatever chains, I don’t think we can or should enforce that user only interact with crypto.org chain.

thomas-nguy: Yeah but the ERC20 wrapping will be only for the native tokens from crypto-org
This transaction is to do the unwrapping and send back to crypto-org.

For ther IBC use cases, we can define additional transactions.

Anyways as I mention, lets just focus on the CRO token flow in this PR.

proto/cronos/cronos.proto
yihuang: I see, if you plan to do ibc tokens coversion in another PR, maybe we can focus on CRO itself in this PR, to avoid some confusion, and easier to review I think.

x/cronos/types/messages.go
JayT106: I see.

x/cronos/keeper/msg_server.go
yihuang: Can you remove the unimplemented part from this PR?

yihuang (changes requested):
The CRO conversion part looks good, can you remove the unimplemented part, the `ConvertEnabled` field? To keep this PR self-contained.

proto/cronos/tx.proto
thomas-nguy: I changed the name to be more generic

x/cronos/keeper/msg_server.go
thomas-nguy: done

x/cronos/keeper/keeper.go
```diff
@@ -24,12 +38,30 @@ func NewKeeper(
...
 	// this line is used by starport scaffolding # ibc/keeper/parameter
 ) *Keeper {
+
+	// set KeyTable if it has not already been set
+	if !paramSpace.HasKeyTable() {
+		paramSpace = paramSpace.WithKeyTable(types.ParamKeyTable())
+	}
+
+	if !evmSpace.HasKeyTable() {
+		paramSpace = paramSpace.WithKeyTable(evmTypes.ParamKeyTable())
```
yihuang, line 54: - `evmSpace` and `paramSpace` don't match?
- Should we initialize EVM param space here? maybe better to get evm params through `EvmKeeper`?

thomas-nguy: yeah for now the Keeper does not need `EvmKeeper` that is why I am only passing the `evmSpace`. You can make this change in your PR if you are planning to add the EvmKeeper to this keeper

x/cronos/types/params.go
```diff
@@ -0,0 +1,59 @@
...
+func NewParams(ibcCroDenom string) Params {
+	return Params{
+		IbcCroDenom: ibcCroDenom,
+	}
+}
+
+// DefaultParams is the default parameter configuration for the cronos module
+func DefaultParams() Params {
+	return Params{
+		IbcCroDenom: "ibc/6B5A664BF0AF4F71B2F0BAA33141E2F1321242FBD5D19762F541EC971ACB0865",
```
yihuang, line 31: Make it a constant variable at the top?

x/cronos/types/params.go
```diff
@@ -0,0 +1,59 @@
...
+	}
+}
+
+// Validate all cronos module parameters
+func (p Params) Validate() error {
+	return validateIsString(p.IbcCroDenom)
```
yihuang, line 37: Validate the `ibc/` prefix and length?

yihuang (approved):
looks good, only minor suggestions.

PR #34, opened 2021-09-01
wrapping native tokens in erc20

### LINKED ISSUE #24: Problem: Wrapping native tokens in evm contract is not supported
Implement the scenario 1 in https://github.com/tharsis/ethermint/discussions/392

- Suggestion from @tomtau : Not allowing gas token to be wrapped in that way, since they can be wrapped normally like `WETH`.

yihuang:
Closes #24

- add dapp tools overlay to nixpkgs
- create CronosERC20 contract
- embed CronosERC20 bytecode in native module
- add keeper methods to transfer tokens between native and contract
- unit test

# PR Checklist:


x/cronos/keeper/evm.go
```diff
@@ -0,0 +1,154 @@
...
+		return nil, err
+	}
+	if res.Failed() {
+		return nil, fmt.Errorf("call contract failed: %s, %s, %s", contract.Hex(), method, res.Ret)
+	}
+	return res.Ret, nil
+}
+
+// DeployCronosERC20 deploy an embed erc20 contract
+func (k Keeper) DeployCronosERC20(ctx sdk.Context, denom string) (common.Address, error) {
```
thomas-nguy, line 74: `DeployModuleContract` to make it more generic? (ERC20 might be changed to CRO20)

x/cronos/keeper/evm.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+
+	if res.Failed() {
+		return common.Address{}, fmt.Errorf("contract deploy failed: %s", res.Ret)
+	}
+	return crypto.CreateAddress(types.EVMModuleAddress, msg.Nonce()), nil
+}
+
+// SendCoinFromNativeToERC20 convert native token to erc20 token
+func (k Keeper) SendCoinFromNativeToERC20(ctx sdk.Context, sender common.Address, coin sdk.Coin) error {
```
thomas-nguy, line 93: `ConvertToEVMTokens` or `ConvertToERC20` to make it more generic?

x/cronos/keeper/evm.go
```diff
@@ -0,0 +1,154 @@
...
+	// TODO validate coin.Denom
+	var err error
+	contract, found := k.GetContractByDenom(ctx, coin.Denom)
+	if !found {
+		contract, err = k.DeployCronosERC20(ctx, coin.Denom)
```
thomas-nguy, line 98: hmm I dont know if we should support this feature

yihuang: if that's the case, it defeats the purpose of most parts of this PR. 😂

thomas-nguy: mm maybe you can just add it for the testing framework?

x/cronos/keeper/evm.go
yihuang: This function is to deploy a specific contract named CronosERC20, so to change the name, we better change that contract name systematically. `CronosToken`?

x/cronos/keeper/evm.go
yihuang: The name `SendCoinFromXToY` is following [bank keeper's style](https://github.com/cosmos/cosmos-sdk/blob/master/x/bank/keeper/keeper.go#L34), the ERC20 represent the ABI, we can replace it with the token standard name we chose in the end.

x/cronos/keeper/evm.go
yihuang: Since we are not sure in the short term, but could need it in the longer term, I added a bool parameter to this method, make the auto-deploy part optional.

x/cronos/keeper/evm.go
thomas-nguy: How about `DeployModuleErc20` or `DeployModuleToken` ?

What bother me is the term `Cronos` because it could refers to the blockchain or the app, not necessarly the the module which is one owning the contract

x/cronos/keeper/evm.go
thomas-nguy: Yeah but this function is not only for `Sending` Coin right?

It does multiple things (sending coin from user account to contract account, deploy contract, mint tokens). I think using the term `Convert` or `Wrapping` would be more accurate

x/cronos/keeper/evm.go
yihuang: Hmm, `ModuleErc20` sounds good, we can change the erc20 to cro20(or whatever) later after we made the decision.

x/cronos/keeper/evm.go
yihuang: Sounds good, changed to `ConvertNativeToERC20/ConvertERC20ToNative`.

yihuang: Oh, we still have the `SendCoin` and `SendCoins` variations, which are not covered by the `Convert*` name, so maybe `ConvertCoinFromNativeToERC20`/`ConvertCoinsFromNativeToERC20`, which is very close to the original one 😄

contracts/src/ModuleERC20.sol
```diff
@@ -0,0 +1,24 @@
...
+
+contract ModuleERC20 is DSToken  {
+    // sha256('cronos')[:20]
+    address constant module_address = 0x6526B48f897f6e08067dF00A1821d783cbC2af8b;
+
+    constructor(string memory denom, uint8 decimals_) DSToken(denom) public {
+        decimals = decimals_;
+    }
+
+    function mint_by_native(address addr, uint amount) public {
```
thomas-nguy, line 13: `mint_by_module` , `burn_by_module`?

Not quite sure of the meaning of native

x/cronos/keeper/evm.go
```diff
@@ -0,0 +1,163 @@
...
+		return err
+	}
+
+	return nil
+}
+
+// ConvertCoinsFromNativeToERC20 convert native tokens to erc20 tokens
+func (k Keeper) ConvertCoinsFromNativeToERC20(ctx sdk.Context, sender common.Address, coins sdk.Coins, autoDeploy bool) error {
+	for _, coin := range coins {
+		if err := k.SendCoinFromNativeToERC20(ctx, sender, coin, autoDeploy); err != nil {
```
thomas-nguy, line 148: rename also `SendCoinFromNativeToERC20` to `ConvertCoinFromNativeToERC20`

x/cronos/keeper/keeper.go
```diff
@@ -29,12 +29,12 @@ type (
...
 		// module specific parameter space that can be configured through governance
 		paramSpace paramtypes.Subspace
-		// evm parameter space
-		evmParamSpace paramtypes.Subspace
 		// update balance and accounting operations with coins
 		bankKeeper types.BankKeeper
 		// ibc transfer operations
 		transferKeeper types.TransferKeeper
+		// ethermint evm keeper
+		evmKeeper *evmkeeper.Keeper
```
thomas-nguy, line 37: would be more clean to create a type.EvmKeeper

x/cronos/keeper/keeper.go
```diff
@@ -215,3 +211,47 @@ func (k Keeper) IbcTransferCoins(ctx sdk.Context, from, destination string, coin
...
+	if len(bz) == 0 {
+		return common.Address{}, false
+	}
+
+	return common.BytesToAddress(bz), true
+}
+
+// GetAutoContractByDenom find the corresponding contract for the denom,
+// external contract is taken in preference to auto-deployed one
+func (k Keeper) GetAutoContractByDenom(ctx sdk.Context, denom string) (common.Address, bool) {
```
thomas-nguy, line 229: I think managing two mapping for autocontract and externalcontract is confusing.

One is enough. The app does not need to manage auto <-> external

yihuang: The reason for the two mappings is when we have auto contract and external contract at the same time for a denom, we still need to support withdrawing from the auto one, while it's possible to cover the withdrawing case without storing the mapping (for example, store the denom in the contract), I feel it's safer to keep that mapping for now, maybe some other use case will need it?

yihuang: Actually, the comment on this function is misleading, fixed.

thomas-nguy: ok but lets try to keep the auto deployment logic confined and lets not make too many special case otherwise it might be confusing to maintain

x/cronos/keeper/keeper.go
```diff
@@ -215,3 +211,46 @@ func (k Keeper) IbcTransferCoins(ctx sdk.Context, from, destination string, coin
...
+		return common.Address{}, false
+	}
+
+	return common.BytesToAddress(bz), true
+}
+
+// GetContractByDenom find the corresponding contract for the denom,
+// external contract is taken in preference to auto-deployed one
+func (k Keeper) GetContractByDenom(ctx sdk.Context, denom string) (contract common.Address, found bool) {
+	contract, found = k.GetExternalContractByDenom(ctx, denom)
```
thomas-nguy, line 241: could `GetExternalContractByDenom` and `GetAutoContractByDenom` be private in that case?

contracts/src/ModuleERC20.sol
yihuang: native means native module, maybe let's keep the full name `mint_by_native_module`.

x/cronos/keeper/keeper.go
yihuang: Yeah, ideally that should be the case, the only obstacle currently is in [this method](https://github.com/crypto-org-chain/cronos/pull/34/files#diff-1cf88287bb640036f7fdeabdc381fba5bd0dc2dc5625d94846fca9359f1f2990R19), it calls many arbitrary methods of `EvmKeeper`, I'd like to finish our part first before pushing for a cleaner interface at upstream, then we change it into an interface.

contracts/src/ModuleERC20.sol
thomas-nguy: or `mint_by_cosmos_module` to be more accurate? I think the word `native` have multiple meaning depending on the context. In the smart contract, native can be the current smart contract or some controllers

x/cronos/keeper/keeper.go
yihuang: done

x/cronos/keeper/keeper.go
yihuang: https://github.com/tharsis/ethermint/issues/533
Opened an issue in ethermint side.

x/cronos/keeper/evm.go
yihuang: By the way, ERC20 -> CRC20 renaming is done too.

contracts/src/ModuleERC20.sol
yihuang: cosmos_module or cronos_module? 😄

x/cronos/keeper/evm.go
yihuang: done

contracts/src/ModuleERC20.sol
thomas-nguy: hmm I dont have strong opinion on that, either is fine as long as we are consistent everywhere?

x/cronos/types/types.go
```diff
@@ -1,8 +1,17 @@
...
 
 var (
 	Ten       = big.NewInt(10)
 	TenPowTen = Ten.Exp(Ten, Ten, nil)
 )
+
+// IsValidDenomToWrap returns if it's ok to wrap the native denom in erc20
+// Currently only supports ibc/{hash} and gravity{contract}
+func IsValidDenomToWrap(denom string) bool {
+	return strings.HasPrefix(denom, "ibc/") || strings.HasPrefix(denom, "gravity0x")
```
thomas-nguy, line 16: types/params.go already have 
`validateIsIbcDenom` 

I think we will also need `validateIsGravityDenom` in the future (when we will add the mapping in genesis)

Maybe we could refactor this method to use both validate

x/cronos/keeper/keeper.go
```diff
@@ -215,3 +211,45 @@ func (k Keeper) IbcTransferCoins(ctx sdk.Context, from, destination string, coin
 	}()
 	return nil
 }
+
+// getExternalContractByDenom find the corresponding external contract for the denom,
+func (k Keeper) getExternalContractByDenom(ctx sdk.Context, denom string) (common.Address, bool) {
```
thomas-nguy, line 216: All the methods below could be move to `params.go`

Are we not keeping the map in the paramSpace so it can be changed through governance?

x/cronos/keeper/evm_test.go
```diff
@@ -0,0 +1,68 @@
...
+	address := common.BytesToAddress(priv.PubKey().Address().Bytes())
+	cosmosAddress := sdk.AccAddress(address.Bytes())
+
+	denom := "ibc/testdenom"
+	amount := big.NewInt(100)
+	coins := sdk.NewCoins(sdk.NewCoin(denom, sdk.NewIntFromBigInt(amount)))
+
+	// mint native tokens
+	err = suite.MintCoins(sdk.AccAddress(address.Bytes()), coins)
+	suite.Require().NoError(err)
```
thomas-nguy, line 35: maybe check also user balance

x/cronos/keeper/evm_test.go
```diff
@@ -0,0 +1,68 @@
...
+
+	// send to erc20
+	err = keeper.ConvertCoinsFromNativeToCRC20(suite.ctx, address, coins, true)
+	suite.Require().NoError(err)
+
+	// check erc20 balance
+	contract, found := keeper.GetContractByDenom(suite.ctx, denom)
+	suite.Require().True(found)
+
+	ret, err := keeper.CallModuleCRC20(suite.ctx, contract, "balanceOf", address)
```
thomas-nguy, line 45: maybe you can check the balance before the call and verify it was 0

x/cronos/keeper/evm.go
```diff
@@ -0,0 +1,169 @@
...
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// ConvertCoinsFromNativeToCRC20 convert native tokens to erc20 tokens
+func (k Keeper) ConvertCoinsFromNativeToCRC20(ctx sdk.Context, sender common.Address, coins sdk.Coins, autoDeploy bool) error {
+	for _, coin := range coins {
```
thomas-nguy, line 153: TODO create cosmos tx for this method and also add some metrics (telemetry or events)

You dont have to do it in this PR

x/cronos/keeper/evm.go
```diff
@@ -0,0 +1,169 @@
...
+}
+
+// ConvertCoinFromCRC20ToNative convert erc20 token to native token
+func (k Keeper) ConvertCoinFromCRC20ToNative(ctx sdk.Context, contract common.Address, receiver common.Address, coin sdk.Coin) error {
+	if !types.IsValidDenomToWrap(coin.Denom) {
+		return errors.New("denom is not supported for wrapping")
+	}
+
+	// validate contract address
+	externalContract, externalFound := k.getExternalContractByDenom(ctx, coin.Denom)
```
thomas-nguy, line 131: just use `GetContractByDenom`

x/cronos/keeper/evm.go
```diff
@@ -0,0 +1,169 @@
...
+	for _, coin := range coins {
+		if err := k.ConvertCoinFromNativeToCRC20(ctx, sender, coin, autoDeploy); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// ConvertCoinsFromCRC20ToNative convert erc20 tokens to native tokens
+func (k Keeper) ConvertCoinsFromCRC20ToNative(ctx sdk.Context, contract common.Address, receiver common.Address, coins sdk.Coins) error {
```
thomas-nguy, line 162: why do you need `contract common.Address` in parameter?

x/cronos/types/types.go
yihuang: done

x/cronos/keeper/evm.go
thomas-nguy, line 162: why do you need `contract common.Address` and `coins sdk.Coins` in parameter?

I think you can only have one coin?

contracts/src/ModuleCRC20.sol
```diff
@@ -0,0 +1,24 @@
+pragma solidity ^0.6.11;
+
+import "ds-token/token.sol";
+
+contract ModuleCRC20 is DSToken  {
```
thomas-nguy, line 5: what is DSToken?

x/cronos/keeper/keeper.go
yihuang: I think with paramSpace, you can only get/set param set as a whole, and the param set is a list of key-value rather than a map.
So I think we need to store in storage, only convert to/from genesis in `InitGenesis` and `ExportGenesis`, that could be done in this issue: https://github.com/crypto-org-chain/cronos/issues/46

x/cronos/keeper/evm.go
yihuang: yeah, it's supposed to be called by some hooks or message handlers.

x/cronos/keeper/keeper.go
thomas-nguy: In bank module, the `SendEnabled` in a map<Coin, boolean> which is governed through governance, I think we can take the same design
https://github.com/cosmos/cosmos-sdk/blob/master/x/bank/types/params.go

x/cronos/keeper/evm.go
yihuang: Here it needs to support both auto and external contracts, user should be able to convert from both.

contracts/src/ModuleCRC20.sol
yihuang: https://github.com/dapphub/ds-token/blob/master/src/token.sol
A base implementation of erc20 token.

x/cronos/keeper/evm.go
yihuang: yeah, `ConvertCoins` don't make sense in this case, I'll remove it.

x/cronos/keeper/evm.go
thomas-nguy: ok I see, maybe in that case u dont need to specify coin but just the amount.

You check if the contract address exists in a reverse map, and you fetch the denom from reverse map.

I am not sure if the denom can be easily found otherwise,  especially if this methods will be called in smart contract.
I imagine smart contract event will be something like 
"__CosmosToNativeToken(amount)" or something like that

x/cronos/keeper/evm_test.go
yihuang: It only calls cosmos-sdk's methods, I think we can trust that? 😄

x/cronos/keeper/evm.go
yihuang: Actually, I've thought about just store the reversed index, but with the current approach, we can avoid that to save the space in storage by storing the denom in the contract. Since the denom is just `gravity{original_contract_address}`, we are just asking the contract to record their own original address, that looks like a reasonable requirement to me.
Something like this:
```solidity
function original_contract() {
     return 0x...
}
function native_denom() {
    return 'gravity' + original_contract();
}
```

x/cronos/keeper/keeper.go
yihuang: It depends on the number of entries in the map, because in param set, complexity of get/set operation is O(N). So unless the number of entries in the map is very low, then it's ok.

contracts/src/ModuleERC20.sol
yihuang: renamed to `mint_by_cronos_module`.

contracts/src/ModuleCRC20.sol
```diff
@@ -0,0 +1,24 @@
...
+
+    function mint_by_cronos_module(address addr, uint amount) public {
+        require(msg.sender == module_address);
+        mint(addr, amount);
+    }
+
+    function burn_by_cronos_module(address addr, uint amount) public {
+        require(msg.sender == module_address);
+        // Deduct user's balance without approval
+        balanceOf[addr] = sub(balanceOf[addr], amount);
```
thomas-nguy, line 21: Do we need to emit event as well?

What happen when the sub fails?

yihuang: > Do we need to emit event as well?

good point, added.

> 
> What happen when the sub fails?

If `sub` fails, the native EVM call fails too, then the message which calls this function fails too.

thomas-nguy: yeah maybe adding 
```
require(balanceOf[address] >= amount, "insufficient-balance");
```

would be more clean in term of users experience (the error message would be more clear?)

x/cronos/keeper/evm.go
thomas-nguy: so how about removing  `coin sdk.Coin` from the parameter? The denom can be computed from the contract address and it would be easier to use

x/cronos/keeper/keeper.go
thomas-nguy: do you mean it will be more efficient or less efficient moving it to param set?

I think we will need an entry in `params.go` anyways in order to change the map through governance

contracts/src/ModuleCRC20.sol
yihuang: 👍 done

x/cronos/keeper/evm.go
yihuang: > so how about removing `coin sdk.Coin` from the parameter? The denom can be computed from the contract address and it would be easier to use and replace by amount

But to query denom here, we have to do an EVM call, it would be more efficient to require the contract to pass the denom and just verify it?

thomas-nguy: I though the denom is gravity + contract_address ? 

Oh you mean original contract address in ethereum. Yeah in that case, we could have a reverse mapping `contract_address_in_cronos -> native_token_denom`

It is not a big issue but more to enhance user experience. The main concern is that native_token_denom is not easy to find for users, usually when you want to convert from crc20 to native, you just care about the contract in cronos, you dont care about the ethereum contract. it might be confusing to use.

thomas-nguy: but if it is too bothersome to do, we can have first the denom recorded in SC as you mentioned. In that case we need to modify auto-deployment logic?

It also make external deployment a bit more complex

yihuang: > I though the denom is gravity + contract_address ?
> 
> Oh you mean original contract address in ethereum. Yeah in that case, we could have a reverse mapping `contract_address_in_cronos -> native_token_denom`
> 
> It is not a big issue but more to enhance user experience. The main concern is that native_token_denom is not easy to find for users, usually when you want to convert from crc20 to native, you just care about the contract in cronos, you dont care about the ethereum contract. it might be confusing to use.

hmm, fair enough.

x/cronos/keeper/keeper.go
yihuang: > do you mean it will be more efficient or less efficient moving it to param set?
> 
> I think we will need an entry in `params.go` anyways in order to change the map through governance

It's much less efficient moving to param set, because when you modify it, it's always load/set as a whole, to find an entry, it's an O(N) scanning.

x/cronos/keeper/evm.go
yihuang: done

thomas-nguy (approved):
LGTM,

The only thing remaining would be to make the denom <-> contract mapping updatable through governance. But I think you can do that in another PR

PR #35, opened 2021-09-01
Add support for ibc hooks & add unit tests

thomas-nguy:
#33 #36

Tested with Solo Machine and it works fine

Todo 
- [x] Add tests
# PR Checklist:

- [x] Have you read the [CONTRIBUTING.md](https://github.com/crypto-org-chain/chain-main/blob/master/CONTRIBUTING.md)?
- [x] Does your PR follow the [C4 patch requirements](https://rfc.zeromq.org/spec:42/C4/#23-patch-requirements)?
- [x] Have you rebased your work on top of the latest master? 
- [x] Have you checked your code compiles? (`make`)
- [ ] Have you included tests for any non-trivial functionality?
- [x] Have you checked your code passes the unit tests? (`make test`)
- [x] Have you checked your code formatting is correct? (`go fmt`)
- [x] Have you checked your basic code style is fine? (`golangci-lint run`)
- [ ] If you added any dependencies, have you checked they do not contain any known vulnerabilities? (`go list -json -m all | nancy sleuth`)
- [ ] If your changes affect the client infrastructure, have you run the integration test?
- [ ] If your changes affect public APIs, does your PR follow the [C4 evolution of public contracts](https://rfc.zeromq.org/spec:42/C4/#26-evolution-of-public-contracts)?
- [ ] If your code changes public APIs, have you incremented the crate version numbers and documented your changes in the [CHANGELOG.md](https://github.com/crypto-org-chain/chain-main/blob/master/CHANGELOG.md)?
- [ ] If you are contributing for the first time, please read the agreement in [CONTRIBUTING.md](https://github.com/crypto-org-chain/chain-main/blob/master/CONTRIBUTING.md) now and add a comment to this pull request stating that your PR is in accordance with the [Developer's Certificate of Origin](https://github.com/crypto-org-chain/chain-main/blob/master/CONTRIBUTING.md#developer-certificate-of-originn).

Thank you for your code, it's appreciated! :)

PR #45, opened 2021-09-03
Problem: evm contract can't call native module message

### LINKED ISSUE #29: Problem: evm contract can't call native module message
- adr to ethermint: https://github.com/tharsis/ethermint/pull/429
- ref impl: https://github.com/tharsis/ethermint/pull/417

With above improvements to ethermint, we can implement EvmHooks in cronos to allow evm contract to call native modules, two potential use cases:
- evm contract call gravity `SendToEth` message, to make gravity bridge a eth-to-eth from user's perspective, make cosmos part become transparent to user.
- evm contract send ibc tokens through ibc channel.
- evm contract call bank send message, I think we don't have an immediate use case for that yet.

yihuang:
Closes #29 

Solution:
- implement bank and gravity evm hooks

# PR Checklist:


x/cronos/keeper/evmhooks.go
```diff
@@ -0,0 +1,148 @@
...
+			Type:    uint256Type,
+			Indexed: false,
+		}, abi.Argument{
+			Name:    "denom",
+			Type:    stringType,
+			Indexed: false,
+		}},
+	)
+	GravitySendEvent = abi.NewEvent(
+		"__CosmosNativeGravitySend",
```
thomas-nguy, line 49: I think it would be better to define more generic event for solidity users that does not know much about cosmos ecosystem. The term "CosmosNative" is quite confusing

Something like "TransferToEthereum" or "SendToChain(chainName,...)"

yihuang: One thing to consider here is to try to prevent name conflict when some contract does not intent to call native functionality but defines identical event name and signatures.

thomas-nguy: yeah maybe we can add a special character or prefix at the beginning to define the list of special instruction that can be uses in smart contract like 

@BankSend()  or CronosBankSend()
@IBCTransfer() or CronosIBCTransfer()
@EthereumTransfer() or CronosEthereumTransfer()

(Not sure if special character is allowed in events)

yihuang: - `__CronosNativeTransfer`
- `__CronosIBCTransfer`
- `__CronosEthereumTransfer`

Let's do this then.

x/cronos/keeper/evmhooks.go
```diff
@@ -0,0 +1,148 @@
...
+
+func NewBankSendHook(bankKeeper types.BankKeeper, cronosKeeper Keeper) *BankSendHook {
+	return &BankSendHook{
+		bankKeeper:   bankKeeper,
+		cronosKeeper: cronosKeeper,
+	}
+}
+
+func (h BankSendHook) PostTxProcessing(ctx sdk.Context, txHash common.Hash, logs []*ethtypes.Log) error {
+	for _, log := range logs {
```
thomas-nguy, line 81: could be cleanup a bit to loop only one time

I think instead of creating 2 hooks (for bank and gravity), you can define another interface which implement EventProcessing

yihuang: Yeah, do you mean to register the log processors into sth like: `event_id -> processer`? sounds good.

calvinaco: Will the log/event order make a difference? And in some cases a processor may need to handle multiple events?

yihuang: I just pushed a new implementation.

> Will the log/event order make a difference? And in some cases a processor may need to handle multiple events?

Currently we only have one handler for one event, if we need to support more, can change the `EventID` method to return an array.

leejw51crypto: where is decimal conversion 
10^8 -> 10^18 ?

yihuang: https://github.com/crypto-org-chain/cronos/pull/35
that is handled in this PR

leejw51crypto (approved):
lgtm

integration_tests/contracts/contracts/TestERC20A.sol
```diff
@@ -2,7 +2,14 @@ pragma solidity ^0.6.6;
...
 contract TestERC20A is ERC20 {
+	event __CronosNativeTransfer(address recipient, uint256 amount);
+	event __CronosEthereumTransfer(address recipient, uint256 amount, uint256 bridge_fee);
+
 	constructor() public ERC20("Bitcoin MAX", "MAX") {
 		_mint(msg.sender, 100000000000000000000000000);
 	}
+
+	function test_native_transfer(uint amount) public {
+		emit __CronosNativeTransfer(msg.sender, amount);
```
thomas-nguy, line 13: Don't u have to burn the amount?

imo `__CronosNativeTransfer` is not clear for the operation. A transfer imply an origin and a destination. 

Here it seems that you are either converting or minting?

It seems more that you are con

x/cronos/keeper/evmhooks.go
```diff
@@ -0,0 +1,191 @@
...
+	}
+	_, err = h.gravitySrv.SendToEthereum(sdk.WrapSDKContext(ctx), &msg)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+// LogProcessEvmHook is an evm hook that convert specific contract logs into native module calls
+type LogProcessEvmHook struct {
```
thomas-nguy, line 161: Can maybe moved to another file to improve visibility

integration_tests/contracts/contracts/TestERC20A.sol
yihuang: This is solely for testing the native call, the semantic of the call is to transfer mapped native token from contract address to recipient.
But because we can't setup token mapping in e2e testing right now, so the test itself is not complete yet (there's a TODO in that).

x/cronos/keeper/evmhooks.go
yihuang: hmm, I moved the log handlers to separated files now.

x/cronos/keeper/evm_log_handlers.go
```diff
@@ -0,0 +1,154 @@
...
+
+	contractAddr := sdk.AccAddress(contract.Bytes())
+	recipient := sdk.AccAddress(unpacked[0].(common.Address).Bytes())
+	amount := sdk.NewIntFromBigInt(unpacked[1].(*big.Int))
+	bridgeFee := sdk.NewIntFromBigInt(unpacked[2].(*big.Int))
+	msg := gravitytypes.MsgSendToEthereum{
+		Sender:            contractAddr.String(),
+		EthereumRecipient: recipient.String(),
+		Amount:            sdk.NewCoin(denom, amount),
+		// FIXME bridge fee?
```
calvinaco, line 146: May I know what this FIXME concerns about?

integration_tests/contracts/contracts/TestERC20A.sol
thomas-nguy: so maybe `__CronosCosmosConvert` or `__CronosCosmosMint`?

x/cronos/keeper/evm_log_handlers.go
```diff
@@ -0,0 +1,154 @@
...
+)
+
+var (
+	_ types.EvmLogHandler = NativeTransferHandler{}
+	_ types.EvmLogHandler = EthereumTransferHandler{}
+)
+
+var (
+	// NativeTransferEvent represent the signature of
+	// `event __CronosNativeTransfer(address recipient, uint256 amount, string denom)`
```
thomas-nguy, line 22: I think you dont need the denom

x/cronos/keeper/evm_log_handlers.go
```diff
@@ -0,0 +1,154 @@
...
+	// EthereumTransferEvent represent the signature of
+	// `event __CronosEthereumTransfer(address recipient, uint256 amount, uint256 bridge_fee)`
+	EthereumTransferEvent abi.Event
+)
+
+func init() {
+	addressType, _ := abi.NewType("address", "", nil)
+	uint256Type, _ := abi.NewType("uint256", "", nil)
+	NativeTransferEvent = abi.NewEvent(
+		"__CronosNativeTransfer",
```
thomas-nguy, line 34: could use const for those string

x/cronos/keeper/evm_log_handlers.go
```diff
@@ -0,0 +1,154 @@
...
+	unpacked, err := NativeTransferEvent.Inputs.Unpack(data)
+	if err != nil {
+		// log and ignore
+		h.cronosKeeper.Logger(ctx).Error("log signature matches but failed to decode", "error", err)
+		return nil
+	}
+
+	denom, found := h.cronosKeeper.GetDenomByContract(ctx, contract)
+	if !found {
+		return errors.New("contract is not connected to native token")
```
thomas-nguy, line 94: I am thinking if we could wrap add those errors in the ethereum log, it could be nice

thomas-nguy: can you add the contract address in the error?

x/cronos/keeper/evm_log_handlers.go
```diff
@@ -0,0 +1,154 @@
...
+	amount := sdk.NewIntFromBigInt(unpacked[1].(*big.Int))
+	bridgeFee := sdk.NewIntFromBigInt(unpacked[2].(*big.Int))
+	msg := gravitytypes.MsgSendToEthereum{
+		Sender:            contractAddr.String(),
+		EthereumRecipient: recipient.String(),
+		Amount:            sdk.NewCoin(denom, amount),
+		// FIXME bridge fee?
+		BridgeFee: sdk.NewCoin(denom, bridgeFee),
+	}
+	_, err = h.gravitySrv.SendToEthereum(sdk.WrapSDKContext(ctx), &msg)
```
thomas-nguy, line 149: would it be more clean to call the keeper instead of msg_server? 

but I dont have strong opinion on this

x/cronos/keeper/evm_log_handlers.go
```diff
@@ -0,0 +1,154 @@
...
+	coins := sdk.NewCoins(sdk.NewCoin(denom, sdk.NewIntFromBigInt(unpacked[1].(*big.Int))))
+	err = h.bankKeeper.SendCoins(ctx, contractAddr, recipient, coins)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// EthereumTransferHandler handles `__CosmosNativeGravitySend` log
```
thomas-nguy, line 108: missing some test cases for this handler

x/cronos/keeper/evm_log_handlers.go
yihuang: The gravity implements the method on the server type rather than the keeper, thus we need the server to call the method.

x/cronos/keeper/evm_log_handlers.go
yihuang: > I am thinking if we could wrap add those errors in the ethereum log, it could be nice

I believe that’s the case, the error returned by hook is saved in the return value of the EVM call.
But I need to double check on the rpc response.

x/cronos/keeper/evm_log_handlers.go
yihuang: It should be removed now 😄

x/cronos/keeper/evm_log_handlers.go
thomas-nguy: you can use `keeper.createSendToEthereum` directly 

The cons is that you are losing the event and validation logic

The pro is that it is easier to mock. You can also define your own event

thomas-nguy: hmm actually the way it is implemented, its not easier to mock... I guess its fine

integration_tests/contracts/contracts/TestERC20A.sol
yihuang: > so maybe `__CronosCosmosConvert` or `__CronosCosmosMint`?

“Native” stands for native token, “Transfer” is also self explanatory I believe?

thomas-nguy: `is also self explanatory I believe?`

I dont think so..
"Native" has barely any meaning without context. And using "Transfer" without specifying a sender and destination sounds weird. 

If we can't find a proper short term to describe the operation, I would prefer to be very explicit and call the event 
"__CronosSendFromContractToAccount(receiver, amount)" instead...

yihuang: > `is also self explanatory I believe?`
> 
> I dont think so..
> "Native" has barely any meaning without context. And using "Transfer" without specifying a sender and destination sounds weird.
> 
> If we can't find a proper short term to describe the operation, I would prefer to be very explicit and call the event
> "__CronosSendFromContractToAccount(receiver, amount)" instead...
> 
> Mind that most smart contract developer won't have deep knowledge about the internal implementation, or cosmos implementation

I think the full version should be `__CronosSendNativeTokensFromContractToAccount` 😄

calvinaco: I agree with @thomas-nguy ... generally the best is we have a concise term, sometimes involve some domain knowledge, for the action; if not better be explicit than ambiguous.

x/cronos/keeper/evm_log_handlers.go
yihuang: fixed

integration_tests/contracts/contracts/TestERC20A.sol
yihuang: How about `__CronosNativeTokenTransfer`, if we look at the signature of the event, it's clear the first argument is the recipient:
```
event __CronosNativeTokenTransfer(address recipient, uint256 amount);
```

Also any suggestion to the name of `__CronosEthereumTransfer`?

x/cronos/keeper/evm_log_handlers.go
yihuang: fixed

x/cronos/keeper/evm_log_handlers.go
yihuang: done

integration_tests/contracts/contracts/TestERC20A.sol
calvinaco: In fact, the name CronosSendNativeTokensFromContractToAccount may still be confusing. Perhaps I overlook sth. But you don't have native tokens in a smart contract, so how can you send native tokens from contract to account? So at the end the name is still not self-explanatory?

But I got a smell that maybe it's over-correct for now 😂 because @yihuang you mentioned this test case is incomplete, so the true meaning may change again when work on it later.

In the code below it seems that it is trying to mint a new native token and credit to the user account?

thomas-nguy: when I read this line 
```__CronosNativeTokenTransfer(msg.sender, amount);```

I interpret it as "I am sending CRO amount from my account to the contract"

I think it is worth to spend some time so that we can have a consensus on the cronos events because they will be a standard for our platform in the future

Comment (thomas-nguy) [reply integration_tests/contracts/contracts/TestERC20A.sol:13]: actually the correct term would be 
``` __CronosSendCoinsFromContractToAccount ``` 

Because bank module use the term coin, not native token,  but since we don't specify the coin denom 

A more accurate term would be 

``` __CronosSendCosmosContractCoinFromContractToAccount ```  but well, its quite long

Comment (yihuang) [reply integration_tests/contracts/contracts/TestERC20A.sol:13]: > In fact, the name CronosSendNativeTokensFromContractToAccount may still be confusing. Perhaps I overlook sth. But you don't have native tokens in a smart contract, so how can you send native tokens from contract to account? So at the end the name is still not self-explanatory?
> 
> But I got a smell that maybe it's over-correct for now 😂 because @yihuang you mentioned this test case is incomplete, so the true meaning may change again when work on it later.
> 
> In the code below it seems that it is trying to mint a new native token and credit to the user account?

The meaning itself is clear here, it's just transformed to the call: `bankKeeper.SendCoins(contract, recipient, sdk.NewCoin(denom, amount))`.

Comment (thomas-nguy) [reply integration_tests/contracts/contracts/TestERC20A.sol:13]: Actually maybe we should change the event to be 
```
__CronosSendCoinFromContractToAccount(recipient, amount, denom)

```

That way, the trigger can be generic. You can send "Any coin" from the contract address to the recipient account, not only the gravity linked coin

File: x/cronos/keeper/evm_log_handlers.go (hunk 8 above)
Comment (yihuang) [reply x/cronos/keeper/evm_log_handlers.go]: https://github.com/tharsis/ethermint/blob/main/x/evm/keeper/state_transition.go#L198

Hmm, the detailed error won't appear in the response, the RPC user only sees a fixed error string.
But it do get logged in a logfile as an error.

File: integration_tests/contracts/contracts/TestERC20A.sol (hunk 3 above)
Comment (yihuang) [reply integration_tests/contracts/contracts/TestERC20A.sol:13]: I guess my habit of using native tokens coming from [cardano](https://developers.cardano.org/docs/native-tokens/).

Comment (calvinaco) [reply integration_tests/contracts/contracts/TestERC20A.sol:13]: > and the contract address should have the native tokens already.
Ohhhh, ok so I think this is the context I have overlooked before. Does it mean every contract will keep its own Cosmos native token?

Comment (yihuang) [reply integration_tests/contracts/contracts/TestERC20A.sol:13]: > > and the contract address should have the native tokens already.
> 
> Ohhhh, ok so I think this is the context I have overlooked. Does it mean every contract will keep its own Cosmos native token?

Yeah, when one converts native tokens to erc20 tokens, those tokens are locked under the contract address.

Comment (calvinaco) [reply integration_tests/contracts/contracts/TestERC20A.sol:13]: > I think it is worth to spend some time so that we can have a consensus on the cronos events because they will be a standard for our platform in the future

I agree, maybe we should have a separate issue for this discussion. That would better record our decision making process.

> Because bank module use the term coin, not native token, but since we don't specify the coin denom

Yes, I think we also need to better define these terms as well. I do use `native` all around to represent token that exists at the Cosmos level. i.e. Coin. But maybe this is not the convention in the Cosmos community. Have to be careful about this.

Comment (yihuang) [reply integration_tests/contracts/contracts/TestERC20A.sol:13]: > I agree, maybe we should have a separate issue for this discussion. That would better record our decision making process.

 https://github.com/crypto-org-chain/cronos/issues/60
 Opened an issue for this.

Comment (calvinaco) [reply integration_tests/contracts/contracts/TestERC20A.sol:13]: Thanks @yihuang 

Shall we discuss event naming pattern in the same issue or a separate one?

Comment (yihuang) [reply integration_tests/contracts/contracts/TestERC20A.sol:13]: `__CronosXTransferFromContract(address recipient, uint256 amount)`

Assume `X` here stands for the cosmos native token denom that is associated with the contract, I guess this pattern is clear enough?

Comment (calvinaco) [reply integration_tests/contracts/contracts/TestERC20A.sol:13]: When I think about it again, can we say "From" is implied because the action is "Send" and we are executing a contract? It is the "Subject" (what token) and "To" that's unclear.

So
```
___CronosSendXToAccount // Account or Native Account? need to address in the terminology issue
___CronosSendXToEthereum // Send X (From the contract account) To Ethereum
___CronosSendXToIBCChannel
// Similar to Gravity Bridge `SendToCosmos` but be more explicit about the token type
```

p.s. Having "From" is no harm, my motive is if we think event name is too long then we can think about to trim it or not. I am okay to include it and be more explicit. (Is there any restricting on event name length?)

x/cronos/keeper/evm_log_handlers.go
yihuang: yeah, just added, one negative case, one positive case.

x/cronos/keeper/evm_log_handlers.go
thomas-nguy: okay I created a new issue for that #61

thomas-nguy (approved):
LGTM except for the event name convention.

integration_tests/contracts/contracts/TestERC20A.sol
yihuang: > When I think about it again, can we say "From" is implied because the action is "Send" and we are executing a contract? It is the "Subject" (what token) and "To" that's unclear.
> 
> So
> 
> ```
> ___CronosSendXToAccount // Account or Native Account? need to address in the terminology issue
> ___CronosSendXToEthereum // Send X (From the contract account) To Ethereum
> ___CronosSendXToIBCChannel
> // Similar to Gravity Bridge `SendToCosmos` but be more explicit about the token type
> ```
> 
> p.s. Having "From" is no harm, my motive is if we think event name is too long then we can think about to trim it or not. I am okay to include it and be more explicit. (Is there any restricting on event name length?)

Looks good to me.
And let's use `NativeToken` for `X` for now?

thomas-nguy: Maybe 'X' can be omitted now?

__CronosSendToAccount
__CronosSendToEthereum
__CronosSendToIBCChannel

Even if we add `NativeToken` or `Coins`, it is not clear which coin is it. We need a proper term to designate the "Coin associated with the contract"

yihuang: > Maybe 'X' can be omitted now?
> 
> __CronosSendToAccount
> __CronosSendToEthereum
> __CronosSendToIBCChannel
> 
> Even if we add `NativeToken` or `Coins`, it is not clear which coin is it. We need a proper term to designate the "Coin associated with the contract"

Looks good too, I suppose we'll have detailed documentation about them as part of the CRC20 standard.

yihuang: https://github.com/crypto-org-chain/cronos/issues/62
opened issue and documented decision so far in it.

PR #149, opened 2021-10-06
Problem: Ledger is not enabled and does not work with Crypto.org Chain Ledger app

### LINKED ISSUE #147: Problem: Ledger is not enabled in the build binary
In the current binary, if one tries to sign with ledger it will return the following error
> support for ledger devices is not available in this executable

### LINKED ISSUE #148: Problem: Default derivation path does not work on Crypto.org Ledger App
The current derivation path  m/44'/60'/0'/0/0 is not supported on Cosmos nor Crypto.org Chain Ledger App.

The more proper solution would be to fork the Cosmos Ledger app and create a Cronos Ledger App with the derivation path, but this will take time to be available on Ledger Live

Given the time constraint, a temp solution for mainnet launch is to build a binary with the Coin Type 394 so that the Crypto.org Chain Ledger App could work with Cronos.

calvinaco:
Solution: (Fix #147) (Fix #148) Add build flag with Crypto.org Chain BIP44 coin type

-----

The use of Coin Type `394` when build with Ledger support is a temp solution so that we have at least the Crypto.org Chain Ledger app to be able to work with `cronosd`.

### DIFF
--- a/cmd/cronosd/config/hdpath_ledger.go
+++ b/cmd/cronosd/config/hdpath_ledger.go
@@ -1,0 +1,13 @@
+// +build ledger
+
+package config
+
+import (
+    ethaccounts "github.com/ethereum/go-ethereum/accounts"
+)
+
+var (
+    BIP44CoinType uint32 = 394
+
+    BIP44HDPath = ethaccounts.DerivationPath{0x80000000 + 44, 0x80000000 + 394, 0x80000000 + 0, 0, 0}.String()
+)
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,7 @@ COVERAGE ?= coverage.txt
 GOPATH ?= $(shell $(GO) env GOPATH)
 BINDIR ?= ~/go/bin
 NETWORK ?= mainnet
+LEDGER_ENABLED ?= false
 
 TESTNET_FLAGS ?=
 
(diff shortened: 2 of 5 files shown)


tomtau (commented):
is there a need to have a build tag for the coin type? one can always pass the command line argument if the non-Ethereum coin type is necessary?

calvinaco:
> is there a need to have a build tag for the coin type? one can always pass the command line argument if the non-Ethereum coin type is necessary?

Yes, that could make this more flexible. I have to research on how to do it. Maybe via `Idflags`?

tomtau:
> Yes, that could make this more flexible. I have to research on how to do it. Maybe via Idflags?

No, I meant that the `cronosd` app can be passed the coin type argument, so no need for two binaries / making a build tag ?

cmd/cronosd/config/hdpath_ledger.go
```diff
@@ -0,0 +1,13 @@
...
+package config
+
+import (
+    ethaccounts "github.com/ethereum/go-ethereum/accounts"
+)
+
+var (
+    BIP44CoinType uint32 = 394
+
+    BIP44HDPath = ethaccounts.DerivationPath{0x80000000 + 44, 0x80000000 + 394, 0x80000000 + 0, 0, 0}.String()
```
yihuang, line 12:
```suggestion
    BIP44HDPath = ethaccounts.DerivationPath{0x80000000 + 44, 0x80000000 + BIP44CoinType, 0x80000000 + 0, 0, 0}.String()
```

yihuang (commented):
The ledger version will generate different addresses? will it cause confusion if the user uses web3 ledger client at the same time?

tomtau:
> The ledger version will generate different addresses? will it cause confusion if the user uses web3 ledger client at the same time?

it will enforce different coin types for different apps -- I think it's ok for CLIs where users are advanced enough to know what they are doing... but it'll be potentially confusing in the end-user GUIs, especially the ones that combine both Cosmos SDK and Ethereum Web3 tooling

calvinaco:
> No, I meant that the cronosd app can be passed the coin type argument, so no need for two binaries / making a build tag ?

Oh I got what you mean now. Yes there's the `--coin-type` flag already. I have reverted the changes regarding the coin type and update to build `Ledger` version by default in Makefile.

leejw51crypto (approved):
lgtm for cro ledger app

tomtau:
that build tag may be redundant -- the only relevant build tag in external dependencies is `ledger_zemu` which shouldn't be enabled by default

yihuang:
> that build tag may be redundant -- the only relevant build tag in external dependencies is `ledger_zemu` which shouldn't be enabled by default

but it seems all this PR does is adding the `leger` build tag? did I miss anything?

tomtau:
> Then I guess it could be enabled by default, it seems the only side effect is extra dependency?
yes -- the only reason for disabling  itmay be in integrations tests where the emulator is used

PR #201, opened 2021-11-03
Problem: no integration test for state sync #6

### LINKED ISSUE #6: Problem: no integration test for state sync
Check state sync behavior in integration test:
- run the devnet
- do some transactions
- add a new state sync node, sync
  - check query chain state works
  - check query old transaction don't work
  - check execute new transaction works
  - check query new transaction works

damoncro:

Close #6
# PR Checklist:

- [x] If your changes affect the client infrastructure, have you run the integration test?

### DIFF
--- a/integration_tests/conftest.py
+++ b/integration_tests/conftest.py
@@ -76,5 +76,5 @@ def cronos_cluster(worker_index, pytestconfig, tmp_path_factory):
         Path(__file__).parent / "../scripts/cronos-devnet.yaml",
         worker_index,
         tmp_path_factory.mktemp("data"),
-        quiet=True,
+        # quiet=True,
     )
--- a/integration_tests/conftest.py
+++ b/integration_tests/conftest.py
@@ -1,6 +1,10 @@
+import re
+from pathlib import Path
+
 import pytest
 
 from .network import setup_cronos, setup_geth
+from .utils import cluster_fixture
 
 
 def pytest_configure(config):
@@ -59,3 +63,18 @@ def cluster(request, cronos, geth):
         yield geth
     else:
         raise NotImplementedError
+
+@pytest.fixture(scope="session")
+def worker_index(worker_id):
+    match = re.search(r"\d+", worker_id)
+    return int(match[0]) if match is not None else 0
+
+@pytest.fixture(scope="session")
+def cronos_cluster(worker_index, pytestconfig, tmp_path_factory):
+    "default cluster fixture"
+    yield from cluster_fixture(
+        Path(__file__).parent / "../scripts/cronos-devnet.yaml",
+        worker_index,
+        tmp_path_factory.mktemp("data"),
+        quiet=True,
+    )
--- a/integration_tests/conftest.py
+++ b/integration_tests/conftest.py
@@ -64,12 +64,12 @@ def cluster(request, cronos, geth):
     else:
         raise NotImplementedError
 
-@pytest.fixture(scope="session")
+@pytest.fixture(scope="module")
 def worker_index(worker_id):
     match = re.search(r"\d+", worker_id)
     return int(match[0]) if match is not None else 0
 
-@pytest.fixture(scope="session")
+@pytest.fixture(scope="module")
 def cronos_cluster(worker_index, pytestconfig, tmp_path_factory):
     "default cluster fixture"
     yield from cluster_fixture(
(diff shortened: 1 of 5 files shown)


integration_tests/conftest.py
```diff
@@ -59,3 +63,18 @@ def cluster(request, cronos, geth):
...
+
+@pytest.fixture(scope="module")
+def worker_index(worker_id):
+    match = re.search(r"\d+", worker_id)
+    return int(match[0]) if match is not None else 0
+
+@pytest.fixture(scope="module")
+def cronos_cluster(worker_index, pytestconfig, tmp_path_factory):
+    "default cluster fixture"
+    yield from cluster_fixture(
```
yihuang, line 75: I think you can use the default `cronos` fixture?

integration_tests/pyproject.toml
```diff
@@ -20,6 +20,7 @@ python-dateutil = "^2.8.1"
 pystarport = { git = "https://github.com/crypto-com/pystarport.git", branch = "main" }
 web3 = "^5.20.1"
 eth-bloom = "^1.0.4"
+pytest-xdist = "1.34.0"
```
yihuang, line 23: we didn't use pytest-xdist to run the integration tests, so don't need the dependency.

integration_tests/utils.py
```diff
@@ -139,15 +139,15 @@ def cluster_fixture(
             ini = chain_data / cluster.SUPERVISOR_CONFIG_FILE
             ini.write_text(
                 re.sub(
-                    r"^command = (.*/)?chain-maind",
-                    "command = chain-maind-inst "
+                    r"^command = (.*/)?cronosd",
```
yihuang, line 142: We should remove this function, it's copied from chain-main, not used in cronos, there's `setup_cronos` function in the `network.py`.

integration_tests/pyproject.toml
damoncro: Why we don't use it? Too clumsy?

integration_tests/utils.py
damoncro: I found it is portable if I create multiple nodes with pytest-xdist. And it works in cronos as well.

damoncro: It is ok to use `setup_cronos`, I can try it to create multiple nodes.

integration_tests/conftest.py
damoncro: It is ok to use `cronos` fixture to create multiple nodes. But I found pytest-xdist is also working alright. Just curious about why we abandon `pytest-xidst` in cronos?

yihuang: `pytest-xidst` is for running tests in parallel to save time, the current test running time in Cronos is ok I think.
several issue with `pytest-xdist` based on previous experience:
- fixtures can't be shared in multiple worker processes, so each work processes need to setup it's own cronos devnet, while only a single one is needed in normal case.
- Not easy to distribute the workload evenly, because different test cases could rely on different setup, so in chain-main we distribute test cases by module, so the spetial devnet only need to set up once in one of the worker processes, but the running times of different test module might vary a lot. But distributing test cases by functions might end up with too many duplicated devnet setups in the worker processes.
- The github action exec env get stressed when there are many parallel executions, results in timeout errors.

integration_tests/pyproject.toml
yihuang: https://github.com/crypto-org-chain/cronos/pull/201#discussion_r741622869

tomtau (commented):
for testing, maybe also need to do a small change in the Go code, otherwise this integration test won't be triggered on this PR?

yihuang:
> for testing, maybe also need to do a small change in the Go code, otherwise this integration test won't be triggered on this PR?

the pytest should pick it up automatically since the function name starts with `test_`.
But many CI checks don't run for this PR, is it because of new contributor?

tomtau:
@yihuang I don't think it's because of a new contributor (it'd display the button to maintainers to execute CI actions) -- it's likely because of this diff filter: https://github.com/crypto-org-chain/cronos/blob/main/.github/workflows/build.yml#L32

yihuang:
> @yihuang I don't think it's because of a new contributor (it'd display the button for maintainers to execute CI actions) -- it's likely because of this diff filter: https://github.com/crypto-org-chain/cronos/blob/main/.github/workflows/build.yml#L32

https://github.com/crypto-org-chain/cronos/blob/main/.github/workflows/test.yml#L24
but there's no filter in the integration test workflow.

damoncro:
Closed and followed with https://github.com/crypto-org-chain/cronos/pull/203

PR #233, opened 2021-11-30
Problem: there are duplicated hard coded informations in integration tests

### LINKED ISSUE #10: Problem: there are duplicated hard coded informations in integration tests
follow up on: https://github.com/crypto-org-chain/cronos/pull/8

The genesis keys and addresses are duplicated and hard coded in different places, better to keep them in single place for easier maintenance in the future.

damoncro:

Close #10
This PR adds the following features:
1. Add a `.env` file which consolidates all keys/mnemonic words/address etc. Scripts, python files, and all yamls can read the `.env` file to expand the keys/mnemonic words/address etc 
2. The yaml expansion job is done by pystarport.

Please merge https://github.com/crypto-com/pystarport/pull/37 first. 

# PR Checklist:


damoncro:
I will fix the lint asap. But it is ok to review first.

thomas-nguy:
there are still linter issues

scripts/devnet.yaml
```diff
@@ -8,14 +9,14 @@ cronos_777-1:
   validators:
     - coins: 1000000000000000000stake,1000000000000000000basetcro
       staked: 1000000000000000000stake
-      mnemonic: visit craft resemble online window solution west chuckle music diesel vital settle comic tribe project blame bulb armed flower region sausage mercy arrive release
+      mnemonic: ${VALIDATOR1_MNEMONIC}
```
yihuang, line 12: Is there a simple way to use this config directly for development after this change?

yihuang: Maybe we can support the expansion feature in `pystarport` first?

yihuang:
For the lint issues, you probably want to configure `black` and `isort` as plugin in your editor.

scripts/devnet.yaml
damoncro: Yes, I already support it in local pystarport. It is ok to summit another PR for it.

damoncro:
@yihuang I just configured it. Should be ok now.

yihuang (commented):
I think we should support the expansion feature in `pystarport` directly, so people can still use `pystarport serve --config scripts/devnet.yaml` in cli directly.

damoncro:
ok, I'll change a little bit. By the way the PR here https://github.com/crypto-com/pystarport/pull/37

yihuang:
> ok, I'll change a little bit. By the way the PR here [crypto-com/pystarport#37](https://github.com/crypto-com/pystarport/pull/37)

If that one is merged, we don't need to do the expansion here right?

damoncro:
@yihuang Yes, but `.env` can not work in our nix setting so far if it is relative path, https://github.com/crypto-com/pystarport/pull/37 only works in cli usage.

I still need to do some modifications on this PR.

yihuang:
> @yihuang Yes, but `.env` can not work in our nix setting so far if it is relative path, [crypto-com/pystarport#37](https://github.com/crypto-com/pystarport/pull/37) only works in cli usage, which I tested.

why?

> 
> I still need to do some modifications on this PR.

damoncro:
the `scripts.nix` would copy the yaml files to nix store, if the .env is relative path, it is under `/nix/store`, of course, the `.env` is not there. But I can avoid the setting by not copying to nix store.

yihuang:
> the `scripts.nix` would copy the yaml files to nix store, if the .env is relative path, it is under `/nix/store`, of course, the `.env` is not there. But I can update `default.nix` and `scripts.nix` by not copying to nix store.

This can be solved by overriding the `dotenv` path with a parameter.

damoncro:
Good idea, I can also try parameterise it.

damoncro:
Now change the expansion work to pystarport. But [The commit ](https://github.com/damoncro/cronos/commit/68675b519992ea39435d934358db6bed7dbff579) uses the test version of pystarport, need to finish and merge https://github.com/crypto-com/pystarport/pull/37, before continuing.

thomas-nguy (approved):

integration_tests/pyproject.toml
```diff
@@ -17,9 +17,11 @@ protobuf = "^3.13.0"
 grpcio = "^1.33.2"
 PyYAML = "^5.3.1"
 python-dateutil = "^2.8.1"
-pystarport = { git = "https://github.com/crypto-com/pystarport.git", branch = "main" }
 web3 = "^5.20.1"
 eth-bloom = "^1.0.4"
+python-dotenv = "^0.19.2"
+deepdiff = "^5.6.0"
```
yihuang, line 23:
```suggestion
```
I think we don't need these two dependencies here anymore?

integration_tests/pyproject.toml
```diff
@@ -17,9 +17,11 @@ protobuf = "^3.13.0"
...
+pystarport = {git = "https://github.com/damoncro/pystarport.git", rev = "feature/expansion"}
```
yihuang, line 24: Should change to crypto-com repo after the corresponding PR merged.

integration_tests/utils.py
```diff
@@ -17,15 +17,18 @@
...
 import yaml
 from dateutil.parser import isoparse
+from dotenv import load_dotenv
 from eth_account import Account
 from hexbytes import HexBytes
 from pystarport import cluster, ledger
 from pystarport.ports import rpc_port
 from web3._utils.transactions import fill_nonce, fill_transaction_defaults
 
+load_dotenv("../scripts/.env")
```
yihuang, line 27:
```suggestion
load_dotenv(Path(__file__).parent.parent / "scripts/.env")
```

nix/default.nix
```diff
@@ -24,10 +24,11 @@ import sources.nixpkgs {
       import ./scripts.nix {
         inherit pkgs;
         config = {
-          chainmain-config = ../scripts/chainmain-devnet.yaml;
-          cronos-config = ../scripts/cronos-devnet.yaml;
-          hermes-config = ../scripts/hermes.toml;
-          geth-genesis = ../scripts/geth-genesis.json;
+          chainmain-config = builtins.toString ../scripts/chainmain-devnet.yaml;
```
yihuang, line 27: Why the `builtins.toString` call?

nix/sources.json
```diff
@@ -73,15 +73,15 @@
         "url_template": "https://github.com/<owner>/<repo>/archive/<rev>.tar.gz"
     },
     "pystarport": {
-        "branch": "main",
+        "branch": "feature/expansion",
         "description": "pystarport setup and run local devnet for cosmos like blockchains",
         "homepage": null,
-        "owner": "crypto-com",
+        "owner": "damoncro",
```
yihuang, line 79: TODO change org after corresponding PR merged.

scripts/.env
```diff
@@ -0,0 +1,10 @@
+export VALIDATOR_KEY='826E479F5385C8C32CD96B0C0ACCDB8CC4FA5CACCC1BE54C1E3AA4D676A6EFF5'
+export COMMUNITY_KEY='5D665FBD2FB40CB8E9849263B04457BA46D5F948972D0FE4C1F19B6B0F243574'
+export PASSWORD='123456'
+export VALIDATOR1_MNEMONIC="visit craft resemble online window solution west chuckle music diesel vital settle comic tribe project blame bulb armed flower region sausage mercy arrive release"
+export VALIDATOR2_MNEMONIC="direct travel shrug hand twice agent sail sell jump phone velvet pilot mango charge usual multiply orient garment bleak virtual action mention panda vast"
+export COMMUNITY_MNEMONIC="notable error gospel wave pair ugly measure elite toddler cost various fly make eye ketchup despair slab throw tribe swarm word fruit into inmate"
+export SIGNER1_MNEMONIC="shed crumble dismiss loyal latin million oblige gesture shrug still oxygen custom remove ribbon disorder palace addict again blanket sad flock consider obey popular"
+export SIGNER2_MNEMONIC="night renew tonight dinner shaft scheme domain oppose echo summer broccoli agent face guitar surface belt veteran siren poem alcohol menu custom crunch index"
+export CRONOS_ADMIN="crc12luku6uxehhak02py4rcz65zu0swh7wjsrw0pp"
```
yihuang, line 9: The address and the private keys are actually derived from the mnemonics, it would be better if we can remove this redundancy, but it's ok regardless.

nix/default.nix
damoncro: It is necessary to get the full path of the file (string), instead of copying to nix store. https://github.com/crypto-org-chain/cronos/blob/main/nix/scripts.nix#L15:  if we use  `./scripts/cronos-devnet.yaml`, then quote it with  `${config.cronos-config}` will copy to nix store. 

Not copying to nix sotre has following benefits:
1. Read `.env` with relative path just under the project `script` folder.
2. In nix-shell environment, if we modify the `cronos-devnet.yaml`, we can call start-cronos directly without exiting the nix-shell and re-enter the nix-shell.

damoncro: Before:
```
#!/nix/store/c9cxq1583a85bsq76q2rbnbiwwp7ygxr-bash-4.4-p23/bin/bash
# rely on environment to provide cronosd
export PATH=/nix/store/f347n055kzsyq30m41x481dpsq0zbip1-python3.8-pystarport-0.2.2/bin:$PATH
source /nix/store/ym2jzjqjmzf7azjkc8m9jd4vy0qjrx5p-dotenv
/nix/store/zhicf2wijlr36315xx1nxq5bhh6q90qs-start-cronos /nix/store/k0mfjv85j0sjbfjj6lisriwq3pkzcsqi-cronos-devnet.yaml $@

```


After:

```
#!/nix/store/c9cxq1583a85bsq76q2rbnbiwwp7ygxr-bash-4.4-p23/bin/bash
# rely on environment to provide cronosd
export PATH=/nix/store/2kfp7bbr155h88zy0kglji3ff33f0p4k-python3.8-pystarport-0.2.2/bin:$PATH
source /nix/store/ym2jzjqjmzf7azjkc8m9jd4vy0qjrx5p-dotenv
/nix/store/zhicf2wijlr36315xx1nxq5bhh6q90qs-start-cronos /Users/damonchen/Documents/cronos/scripts/cronos-devnet.yaml $@

```

nix/sources.json
damoncro: Done

integration_tests/utils.py
damoncro: Done

integration_tests/pyproject.toml
damoncro: Done

integration_tests/pyproject.toml
damoncro: Done

scripts/.env
damoncro: I see, is it better to follow in another issue?

damoncro: Created in https://github.com/crypto-org-chain/cronos/issues/240

nix/default.nix
yihuang: Then you don't get the reproducibility with these scripts.
In development, you can always call `./scripts/start-cronos config data` explicitly.

integration_tests/pyproject.toml
damoncro: `python-dotenv` is used in `integration_tests/utils.py`, should be kept.

integration_tests/pyproject.toml
damoncro: `python-dotenv` is used in `integration_tests/utils.py`, should be kept.

nix/default.nix
damoncro: I see, treat it like a fixed binary. Let me think a more thoughtful idea.

damoncro: One question here: since `config` is fixed, only data directory can be changed. Should be `./scripts/start-cronos data` can be called, rather than `./scripts/start-cronos config data`?

damoncro:
```
[nix-shell:~/Documents/cronos/integration_tests]$ start-cronos
No data directory supplied
```

damoncro: Oh, seems you make it wrong? `./scripts/start-cronos config data` is different than `start-cronos` binary. If user calls `./scripts/start-cronos config data`, it is always reproducible.

damoncro: It is ok to make `start-cronos` also reproducible as well.

damoncro: But seems no documents here, user would not know `start-cronos` exists. The binary is not useful because it only works in `nix-shell`. Even me, I don't use it at all.

damoncro: Of course, it is important in integration test. Just not that useful for users, assuming they do not have `nix` installed.

yihuang: `start-cronos` and `start-geth` are only used to start cronos and geth devnets with the identical environment, to run cronos devnet with custom config, one can always use `pystarport` directly, the `scripts/start-cronos` is only a thin wrapper over that.

damoncro: Two TODOs here:
1. Make all `ShellScriptBin` reproducible. And add documents if possible.
2. Improve all scripts. E.g. So that user knows they can use `dotenv` option as well.

integration_tests/pyproject.toml
```diff
@@ -17,9 +17,10 @@ protobuf = "^3.13.0"
 grpcio = "^1.33.2"
 PyYAML = "^5.3.1"
 python-dateutil = "^2.8.1"
-pystarport = { git = "https://github.com/crypto-com/pystarport.git", branch = "main" }
 web3 = "^5.20.1"
 eth-bloom = "^1.0.4"
+python-dotenv = "^0.19.2"
+pystarport = {git = "https://github.com/crypto-com/pystarport.git", rev = "main"}
```
yihuang, line 23:
```suggestion
pystarport = "^0.2.3"
```
`0.2.3` is released on pypi now,

nix/default.nix
yihuang: We can remove `start-chainmain` and `start-cronos`, and simply use `pystarport` directly.

yihuang (approved):

PR #239, opened 2021-12-03
Problem: gravity module is not up to date and pointed to a forked version

thomas-nguy:
Solution: Point to upstream and use latest version v0.2.23 since the bugfix https://github.com/PeggyJV/gravity-bridge/pull/182 has been merged to upstream


tomtau (approved):
👍probably also re-generate the gomod2nix.toml for integration tests?

PR #244, opened 2021-12-07
Problem: The address and the private keys should be derived from the mnemonics in integration test

### LINKED ISSUE #240: Problem: The address and the private keys should be derived from the mnemonics in integration test
**Describe the bug**

**To Reproduce**

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

damoncro:

Close #240 

# PR Checklist:

### DIFF
--- a/integration_tests/utils.py
+++ b/integration_tests/utils.py
@@ -30,7 +30,7 @@ ACCOUNTS = {
     "validator": Account.from_mnemonic(os.getenv("VALIDATOR1_MNEMONIC")),
     "community": Account.from_mnemonic(os.getenv("COMMUNITY_MNEMONIC")),
 }
-KEYS = {name: account.key.hex().replace("0x", "") for name, account in ACCOUNTS.items()}
+KEYS = {name: account.key for name, account in ACCOUNTS.items()}
 ADDRS = {name: account.address for name, account in ACCOUNTS.items()}
 CRONOS_ADDRESS_PREFIX = "crc"
 
--- a/integration_tests/utils.py
+++ b/integration_tests/utils.py
@@ -25,11 +25,13 @@ from pystarport.ports import rpc_port
 from web3._utils.transactions import fill_nonce, fill_transaction_defaults
 
 load_dotenv(Path(__file__).parent.parent / "scripts/.env")
-KEYS = {
-    "validator": os.getenv("VALIDATOR_KEY"),
-    "community": os.getenv("COMMUNITY_KEY"),
+Account.enable_unaudited_hdwallet_features()
+ACCOUNTS = {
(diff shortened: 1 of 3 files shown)


integration_tests/utils.py
```diff
@@ -25,11 +25,13 @@
...
-KEYS = {
-    "validator": os.getenv("VALIDATOR_KEY"),
-    "community": os.getenv("COMMUNITY_KEY"),
+Account.enable_unaudited_hdwallet_features()
+ACCOUNTS = {
+    "validator": Account.from_mnemonic(os.getenv("VALIDATOR1_MNEMONIC")),
+    "community": Account.from_mnemonic(os.getenv("COMMUNITY_MNEMONIC")),
 }
-ADDRS = {name: Account.from_key(key).address for name, key in KEYS.items()}
+KEYS = {name: account.key.hex().replace("0x", "") for name, account in ACCOUNTS.items()}
```
yihuang, line 33:
```suggestion
KEYS = {name: binascii.hexlify(account.key) for name, account in ACCOUNTS.items()}
```

scripts/start-geth
```diff
@@ -24,14 +24,26 @@ $PASSWORD
 EOF
 
 # import validator key
+validator_key=$(python -c """
```
yihuang, line 27: But can we run `start-geth` directly from cli after this?

damoncro: Yes, we can, in nix-shell.

yihuang: where is the `.env` sourced?

yihuang: Or one should source it by himself, or put it in shell profile?

integration_tests/utils.py
damoncro: `hexlify` returns a bytes, should we need it be string?

scripts/start-geth
damoncro: Yes, in nix shell, source in `start-geth` defined in `script.nix`. For the script itself, no source.

damoncro: It is difficult to add it in script, because the `.env` relative path depends on the user's working directory.

damoncro: Or use full path, but full path also can be changed.

integration_tests/utils.py
yihuang: I think `account.key` alone should be good here, in the end, it's passed to `Account.from_key`, which accepts `bytes`.

damoncro: You are right, updated. But it it still necessary in `start-geth` which expects string.

yihuang (approved):

PR #252, opened 2021-12-14
Problem: no integration test for transaction index

### LINKED ISSUE #246: Problem: no integration test for transaction index
There are some complex logics around transaction index, it's good to check them in the integration test:
- Try different status of tx, and check how they are handled by json rpc API
  - failed, check error message, check it's not included in getBlock API.
  - reverted, check it is included in getBlock API, check getBlockAndTxIndex API works ok, check the tx index in the logs are correct.

damoncro:

Close #246 

# PR Checklist:

### DIFF
--- a/integration_tests/test_basic.py
+++ b/integration_tests/test_basic.py
@@ -233,7 +233,7 @@ def test_transaction(cronos):
     assert tx1["transactionIndex"] == 0
 
     # out of gas
-    try:
+    with pytest.raises(ValueError) as error:
         send_transaction(
             w3,
             {
@@ -244,11 +244,10 @@ def test_transaction(cronos):
             },
             KEYS["validator"],
         )["transactionHash"]
-    except ValueError as error:
-        assert "out of gas" in str(error)
+    assert "out of gas" in str(error)
 
     # insufficient fee
-    try:
+    with pytest.raises(ValueError) as error:
         send_transaction(
             w3,
             {
@@ -258,8 +257,7 @@ def test_transaction(cronos):
             },
             KEYS["validator"],
         )["transactionHash"]
-    except ValueError as error:
-        assert "insufficient fee" in str(error)
+    assert "insufficient fee" in str(error)
 
     # Deploy contract
     test_revert = TestRevert(KEYS["validator"])
--- a/integration_tests/test_basic.py
+++ b/integration_tests/test_basic.py
@@ -238,6 +238,36 @@ def test_transaction(cronos):
     tx1 = w3.eth.get_transaction(txhash_1)
     assert tx1["transactionIndex"] == 0
 
+    initial_block_number = w3.eth.get_block_number()
+
+    # tx already in mempool
+    with pytest.raises(ValueError) as exc:
+        send_transaction(
+            w3,
+            {
+                "to": ADDRS["community"],
+                "value": 10000,
+                "gasPrice": w3.eth.gas_price,
+                "nonce": w3.eth.get_transaction_count(ADDRS["validator"]) - 1,
+            },
+            KEYS["validator"],
+        )
+    assert "tx already in mempool" in str(exc)
+
+    # invalid sequence
+    with pytest.raises(ValueError) as exc:
+        send_transaction(
+            w3,
+            {
+                "to": ADDRS["community"],
+                "value": 10000,
+                "gasPrice": w3.eth.gas_price,
+                "nonce": w3.eth.get_transaction_count(ADDRS["validator"]) + 1,
+            },
+            KEYS["validator"],
+        )
+    assert "invalid sequence" in str(exc)
+
     # out of gas
     with pytest.raises(ValueError) as exc:
         send_transaction(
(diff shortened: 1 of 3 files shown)


integration_tests/test_basic.py
```diff
@@ -217,3 +218,70 @@ def test_statesync(cronos):
...
+            {
+                "to": ADDRS["community"],
+                "value": 10000,
+                "gasPrice": w3.eth.gas_price,
+                "gas": 1,
+            },
+            KEYS["validator"],
+        )["transactionHash"]
+    except ValueError as error:
+        assert "out of gas" in str(error)
```
yihuang, line 248: https://docs.pytest.org/en/6.2.x/assert.html#assertions-about-expected-exceptions
there's `pytest.raises` to expect error.

damoncro: But I need to assert the string.

damoncro: Oh raises also ok to assert inside?

damoncro: I'll check

integration_tests/utils.py
```diff
@@ -441,3 +441,83 @@ def call_contact(self, w3):
         assert "world" == self.contract.functions.greet().call()
 
         return send_store_tx
+
+
+class TestRevert:
```
yihuang, line 446: It seems many codes can be reused with `class Greeter`.

damoncro: Yes. This can be improved...

integration_tests/test_basic.py
yihuang: yes, `with pytest.raises(...) as exc_info` and assert the `exc_info`.

yihuang:
- try to include multiple txs in same block, and check the indexes.
- include a cosmos level failure case, and check it doesn't appear in `get_block` response.
- check the tx_index and log_index in tx receipt are correct.
- check `get_transaction_by_block` works.

integration_tests/utils.py
```diff
@@ -109,6 +111,12 @@ def wait_for_ipc(path, timeout=40.0):
             )
 
 
+def wait_for_current_block_confirmation(w3):
```
yihuang, line 114: This method seems like same as `wait_for_new_blocks(1)`.

yihuang (approved):

damoncro:
Sorry, the PR is still under polished. Please hold it first.

integration_tests/utils.py
```diff
@@ -383,61 +388,143 @@ def __init__(self, private_key=KEYS["validator"], chain_id=777):
...
             )
 
-            signed_txn = w3.eth.account.sign_transaction(
-                transaction, private_key=self.private_key
-            )
-
-            # print("Deploying contract...")
-            # Send this signed transaction
-            self._contract_hash = w3.eth.send_raw_transaction(signed_txn.rawTransaction)
+            tx_hash, tx_receipt = self.sign(transaction).send().wait()
```
yihuang, line 404: I think we can use `send_transaction(tx)` here?

integration_tests/utils.py
damoncro: It is different, sometimes the block number did not increase when the first time to call ` w3.eth.get_block_number()`. The purpose here is not simply wait for one block.

integration_tests/utils.py
damoncro: let me replace with send_transaction.

integration_tests/utils.py
yihuang: I don't see the difference, isn't the code is just waiting for a new block mined?

damoncro: Sometimes, I found deploy contracts thread and call contracts thread will make all transactions included in one block. With this, it make sure they will split it into two or three blocks (1 empty).

But now, I can not find it is necessary anymore, I already deleted the `wait_for_current_block_confirmation`.

integration_tests/test_basic.py
```diff
@@ -217,3 +223,176 @@ def test_statesync(cronos):
...
+
+
+def assert_receipt_transaction_and_block(w3, futures):
+    receipts = []
+    for future in concurrent.futures.as_completed(futures):
+        # name = future_to_contract[future]
+        try:
+            data = future.result()
+        except Exception as exc:
+            print("%s" % (exc))
```
yihuang, line 363: why ignore the exception here?

integration_tests/test_basic.py
```diff
@@ -217,3 +223,176 @@ def test_statesync(cronos):
...
+    for receipt in receipts:
+        # check in the same block
+        assert receipt["blockNumber"] == block_number
+        # check transactionIndex
+        transaction_index = receipt["transactionIndex"]
+        assert transaction_index in tx_indexes
+        tx_indexes.remove(transaction_index)
+
+    block = w3.eth.get_block(block_number)
+    print(block)
```
yihuang, line 380: remove debug prints?

yihuang (approved):
LGTM, just minor suggestions:
- exception handling
- cleanup debug prints

Issue #285, opened 2021-12-31
Problem: contract suicide is not tested


yihuang:
contract suicide compliance test.
- within the tx, after the contract suicided, the code still exists.
- after the tx, the code no longer exist.

PR #286, opened 2021-12-31
Problem: contract suicide is not tested

yihuang:
Closes: #285
Solution:
test that:
- within the tx, after the contract suicided, the code still exists.
- after the tx, the code no longer exist.

BTW, this test case don't pass for current ethermint, it's only fixed in statedb-refactor branch.

# PR Checklist:

### DIFF
--- a/integration_tests/test_basic.py
+++ b/integration_tests/test_basic.py
@@ -448,7 +448,9 @@ def test_suicide(cluster):
     assert len(w3.eth.get_code(destroyee.address)) > 0
     assert len(w3.eth.get_code(destroyer.address)) > 0
 
-    tx = destroyer.functions.codesize_after_suicide(destroyee.address).buildTransaction()
+    tx = destroyer.functions.check_codesize_after_suicide(
+        destroyee.address
+    ).buildTransaction()
     receipt = send_transaction(w3, tx)
     assert receipt.status == 1
 
--- a/integration_tests/test_basic.py
+++ b/integration_tests/test_basic.py
@@ -437,19 +437,19 @@ def test_suicide(cluster):
     - after the tx, the code is not available.
     """
     w3 = cluster.w3
-    inner = deploy_contract(
+    destroyee = deploy_contract(
         w3,
-        contract_path("Inner", "TestSuicide.sol"),
+        contract_path("Destroyee", "TestSuicide.sol"),
     )
-    outer = deploy_contract(
+    destroyer = deploy_contract(
         w3,
-        contract_path("Outer", "TestSuicide.sol"),
+        contract_path("Destroyer", "TestSuicide.sol"),
     )
-    assert len(w3.eth.get_code(inner.address)) > 0
-    assert len(w3.eth.get_code(outer.address)) > 0
+    assert len(w3.eth.get_code(destroyee.address)) > 0
+    assert len(w3.eth.get_code(destroyer.address)) > 0
 
-    tx = outer.functions.codesize_after_suicide(inner.address).buildTransaction()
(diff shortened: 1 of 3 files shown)


tomtau (commented):
probably good to link the reference for how the contract suicide should happen (in the yellow paper?)

yihuang:
> probably good to link the reference for how the contract suicide should happen (in the yellow paper?)

the same `test_suicide` case will run on both geth and cronos (because of the parametrized `cluster` fixture), to check identical behaviors on both chains.

integration_tests/test_basic.py
```diff
@@ -427,3 +428,28 @@ def test_message_call(cronos):
...
+
+
+def test_suicide(cluster):
+    """
+    test compliance of contract suicide
+    - within the tx, after contract suicide, the code is still available.
+    - after the tx, the code is not available.
+    """
+    w3 = cluster.w3
+    inner = deploy_contract(
```
damoncro, line 440: what inner and outer means?

yihuang: they are just the contract names: https://github.com/crypto-org-chain/cronos/pull/286/files#diff-691042267b4f7caaa41185a26ffad9fae48b8960f4c668bba805f4cb782f527fR4
inner contract is to be destroyed by the outer contract.

tomtau: I assumed that after reading through it, but was puzzled by the names as well 😂 perhaps a comment there or renaming it as "to_be_destroyed"/"destroyee" and "destroyer" contracts?

yihuang: sounds good, renamed.

yihuang:
> probably good to link the reference for how the contract suicide should happen (in the yellow paper?)

actually, I didn't find the very specification about this on yellow paper.

adu-crypto:
I do find some descriptions related to self-destructs that imply the code of account should be cleared after tx execution:
```
The tuple contents include As, the self-destruct set: a
set of accounts that will be discarded following the transaction’s completion
```

adu-crypto:
https://github.com/ethereum/go-ethereum/blob/443afc975c7a4016bf9ed1b057b8cc186cb089e1/core/state/statedb.go#L433
```
// Suicide marks the given account as suicided.
// This clears the account balance.
//
// The account's state object is still available until the state is committed,
// getStateObject will return a non-nil account after Suicide.
```
so before the commitment, the code size should not be zero as the test suggested.

yihuang:
> I do find some descriptions related to self-destructs that imply the code of account should be cleared after tx execution:
> 
> ```
> The tuple contents include As, the self-destruct set: a
> set of accounts that will be discarded following the transaction’s completion
> ```

Did you find spec about the behavior between the selfdestruct and tx commit?

adu-crypto:
I find this:
Screen Shot 2022-01-04 at 3 49 29 PM

it says:
```
halt execution and register account for later deletion
```

adu-crypto:
go-ethereum just append suicideChange message to the journal and mark stateObject as suicided.
https://github.com/ethereum/go-ethereum/blob/443afc975c7a4016bf9ed1b057b8cc186cb089e1/core/state/statedb.go#L440
I guess any changes to statedb should not take effect until journal commitment?

yihuang:
> go-ethereum just append suicideChange message to the journal and mark stateObject as suicided. https://github.com/ethereum/go-ethereum/blob/443afc975c7a4016bf9ed1b057b8cc186cb089e1/core/state/statedb.go#L440 I guess any changes to statedb should not take effect until journal commitment?

it's marked as suicided first, only be actually deleted when commit.

adu-crypto:
> 

yes

tomtau (approved):

PR #337, opened 2022-02-13
Problem: miss the exact crc20 token contract deployed

CeruleanAtMonaco:
Solution: add the exact crc20 token contract to match the commit, functionally the same as ModuleCRC20, just the name is set when initiating the instance.


yihuang:
how about change `ModuleCRC20` directly to add the name parameter, that way we only have one contract to maintain.

thomas-nguy:
should we add more comment in the contract to define its purpose?

may also consider opening a specific github repo or a folder for CRC standard and move all the documentation there

PR #382, opened 2022-03-08
Problem: Missing license on contracts folder

calvinaco:
Solution: Add license file


thomas-nguy:
should we replace this one instead?
https://github.com/crypto-org-chain/cronos/blob/main/LICENSE

yihuang (approved):

Issue #431, opened 2022-04-19
Problem: get tx receipt api is too slow for indexing service


yihuang:
get tx receipt api is too slow for indexing service. one request could take more than 100ms, when there are dozens of transactions, total time could be seconds, exceeds the block time.
One solution is to add an api to fetch receipts in batch for a block.

PR #432, opened 2022-04-19
Problem: get tx receipt api is too slow for indexing service

yihuang:
Closes: #431
Solution:
- add extension api: cronos_getTransactionReceiptsByBlock to fetch receipts in batch.

Example:
- add `cronos` namespace to `json-rpc.api` config.
- run:
```
$ curl -X POST --data '{"jsonrpc":"2.0","method":"cronos_getTransactionReceiptsByBlock","params":["0x10"],"id":1}' -H "Content-Type: application/json" $ETH_RPC_URL
```

# PR Checklist:

### DIFF
--- a/integration_tests/test_basic.py
+++ b/integration_tests/test_basic.py
@@ -1,6 +1,5 @@
 import concurrent.futures
 import json
-import tempfile
 import time
 from pathlib import Path
 
@@ -17,9 +16,11 @@ from .utils import (
     KEYS,
     Greeter,
     RevertTestContract,
+    build_batch_tx,
     contract_address,
     contract_path,
     deploy_contract,
+    get_receipts_by_block,
     modify_command_in_supervisor_config,
     send_transaction,
     sign_transaction,
@@ -464,6 +465,7 @@ def test_suicide(cluster):
 def test_batch_tx(cronos):
     "send multiple eth txs in single cosmos tx"
     w3 = cronos.w3
+    cli = cronos.cosmos_cli()
     sender = ADDRS["validator"]
     recipient = ADDRS["community"]
     nonce = w3.eth.get_transaction_count(sender)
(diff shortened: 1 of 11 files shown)


x/cronos/rpc/api.go
```diff
@@ -0,0 +1,238 @@
...
+		clientCtx:    clientCtx,
+		queryClient:  rpctypes.NewQueryClient(clientCtx),
+		chainIDEpoch: eip155ChainID,
+		logger:       logger.With("client", "json-rpc"),
+		backend:      backend,
+	}
+}
+
+// GetTransactionReceiptsByBlock returns all the transaction receipts included in the block.
+func (api *CronosAPI) GetTransactionReceiptsByBlock(blockNrOrHash rpctypes.BlockNumberOrHash) ([]map[string]interface{}, error) {
```
JayT106, line 81: Do we have timeout/cancel context control for running the CronosAPI? Not sure how long to loop the tx result will be.

yihuang: The json-rpc server has read/write timeout, afaik, that will control the timeout of handler execution.

x/cronos/rpc/api.go
```diff
@@ -0,0 +1,238 @@
...
+				"blockHash":        blockHash,
+				"blockNumber":      hexutil.Uint64(blockNumber),
+				"transactionIndex": hexutil.Uint64(txIndex),
+
+				// sender and receiver (contract or EOA) addreses
+				"from": from,
+				"to":   txData.GetTo(),
+			}
+
+			// If the ContractAddress is 20 0x0 bytes, assume it is not a contract creation
```
yihuang, line 202:
```suggestion
			// If the to is empty, assume it is not a contract creation
```

integration_tests/test_basic.py
```diff
@@ -562,8 +524,61 @@ def test_batch_tx(cronos):
...
+    )
+    rsp = cli.broadcast_tx_json(cosmos_tx)
+    assert rsp["code"] == 0, rsp["raw_log"]
+
+    receipts = [w3.eth.wait_for_transaction_receipt(h) for h in tx_hashes]
+    assert receipts[0].status == receipts[1].status == 1
+    assert receipts[2].status == 0
+
+    # test the cronos_getTransactionReceiptsByBlock api
+    rsp = get_receipts_by_block(w3, receipts[0].blockNumber)
```
yihuang, line 568: the new api is called here, and the result is checked against the normal `eth_getTransactionReceipt` response.

x/cronos/rpc/utils.go
```diff
@@ -0,0 +1,59 @@
...
+				case evmtypes.AttributeKeyTxIndex:
+					msg.TxIndex, err = strconv.ParseUint(string(attr.Value), 10, 64)
+					if err != nil {
+						return nil, err
+					}
+				case evmtypes.AttributeKeyEthereumTxFailed:
+					msg.Failed = true
+				}
+			}
+		} else if event.Type == evmtypes.EventTypeTxLog {
```
JayT106, line 48: Does every EventTxLog always come after an EventEthereumTX? Otherwise, how do we guarantee the `msg.Logs` is matched with the correct EthereumTx.

x/cronos/rpc/api.go
crypto-facs: Correct.. this is manage on the Ethermint level

x/cronos/rpc/api.go
```diff
@@ -0,0 +1,238 @@
...
+		return nil, err
+	}
+
+	resBlock, err := api.clientCtx.Client.Block(api.ctx, blockNum.TmHeight())
+	if err != nil {
+		api.logger.Debug("block not found", "height", blockNum, "error", err.Error())
+		return nil, nil
+	}
+
+	blockNumber := resBlock.Block.Height
```
crypto-facs, line 97: We already had the blockNumber on line 86

crypto-facs (approved):
I think this looks good.. I wonder if we could have a common function between this endpoint and `eth_getTransactionReceipt` for ParseEvents especially to maintain synchronization between this endpoint's response and ethermint's endpoint.

x/cronos/rpc/utils.go
yihuang: According to ethermint implementation, they are emitted together.

x/cronos/rpc/api.go
yihuang: there's a difference though, the `blockNum` has the special values for latest/pending, but the `resBlock.Block.Height` is a exact number, I've changed the following logic to use this one.

x/cronos/rpc/utils.go
JayT106: I see. just thinking if we are able to check the txIndex (if both event type has it). So we can make sure the parsing is correct.

yihuang: no, not all event types contains tx index, they are only different fragments of the same tx.

JayT106 (approved):
LGTM

PR #492, opened 2022-05-19
Problem: mempool behaviors are not tested (fix #307)

### LINKED ISSUE #307: Problem: mempool behaviors are not tested
Can configure a long `timeout_commit` to have a longer block time, easier to test mempool.

- `getPendingTransactions`
- `getTransactionsCount`
- pending tx filter

linfeng-crypto:
add transaction and contract tx mempool behave test

### DIFF
--- a/integration_tests/configs/long_timeout_commit.yaml
+++ b/integration_tests/configs/long_timeout_commit.yaml
@@ -63,4 +63,4 @@ cronos_777-1:
       feemarket:
         params:
           no_base_fee: false
-          base_fee: 100000000000
+          base_fee: "100000000000"
--- a/integration_tests/configs/long_timeout_commit.yaml
+++ b/integration_tests/configs/long_timeout_commit.yaml
@@ -1,0 +1,66 @@
+dotenv: ../../scripts/.env
+cronos_777-1:
+  cmd: cronosd
+  start-flags: "--trace"
+  app-config:
+    minimum-gas-prices: 0basetcro
+    index-events:
+      - ethereum_tx.ethereumTxHash
+    json-rpc:
+      address: "0.0.0.0:{EVMRPC_PORT}"
+      ws-address: "0.0.0.0:{EVMRPC_PORT_WS}"
+      api: "eth,net,web3,debug,cronos"
+  validators:
+    - coins: 1000000000000000000stake,10000000000000000000000basetcro
+      staked: 1000000000000000000stake
+      mnemonic: ${VALIDATOR1_MNEMONIC}
+      config:
+        consensus:
+          timeout_commit: 15s
+    - coins: 1000000000000000000stake,10000000000000000000000basetcro
+      staked: 1000000000000000000stake
+      mnemonic: ${VALIDATOR2_MNEMONIC}
+      config:
+        consensus:
+          timeout_commit: 15s
+  accounts:
+    - name: community
+      coins: 10000000000000000000000basetcro
+      mnemonic: ${COMMUNITY_MNEMONIC}
+    - name: signer1
+      coins: 20000000000000000000000basetcro
+      mnemonic: ${SIGNER1_MNEMONIC}
+    - name: signer2
+      coins: 30000000000000000000000basetcro
+      mnemonic: ${SIGNER2_MNEMONIC}
+
+  genesis:
+    consensus_params:
+      block:
+        max_bytes: "1048576"
+        max_gas: "81500000"
+    app_state:
+      evm:
+        params:
+          evm_denom: basetcro
+      cronos:
+        params:
+          cronos_admin: ${CRONOS_ADMIN}
+          enable_auto_deployment: true
+          ibc_cro_denom: ${IBC_CRO_DENOM}
+      gov:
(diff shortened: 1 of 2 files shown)


integration_tests/test_mempool.py
```diff
@@ -0,0 +1,87 @@
...
+    wait_for_new_blocks(cli, 1)
+    block_num_2 = w3.eth.get_block_number()
+    print(f"block number contract begin at height: {block_num_2}")
+    contract = deploy_contract(w3, CONTRACTS["Greeter"])
+    tx = contract.functions.setGreeting("world").buildTransaction()
+    signed = sign_transaction(w3, tx, key_from)
+    txhash = w3.eth.send_raw_transaction(signed.rawTransaction)
+    w3.eth.wait_for_transaction_receipt(txhash)
+    # check tx in mempool
+    new_txs = filter.get_new_entries()
```
yihuang, line 47: to test the mempool, we should try to call this before the tx is included, move this before the `wait_for_transaction_receipt`?

integration_tests/test_mempool.py
```diff
@@ -0,0 +1,87 @@
...
+    assert len(all_pending) == 0
+
+    # check transaction
+    block_num_0 = w3.eth.get_block_number()
+    print(f"block number start: {block_num_0}")
+    nonce_begin = w3.eth.get_transaction_count(address_from)
+
+    sended_hash_list = []
+    for i in range(5):
+        nonce = nonce_begin + i
```
yihuang, line 67:
```suggestion
        nonce = w3.eth.get_transaction_count(address_from)
```

we need to do this to test the API behavior for the pending transactions.

integration_tests/test_mempool.py
```diff
@@ -0,0 +1,87 @@
...
+            "nonce": nonce,
+        }
+        signed = sign_transaction(w3, tx, key_from)
+        txhash = w3.eth.send_raw_transaction(signed.rawTransaction)
+        sended_hash_list.append(txhash)
+    block_num_1 = w3.eth.get_block_number()
+    assert block_num_1 == block_num_0
+    print(f"all send tx hash: f{sended_hash_list}")
+    all_pending = w3.eth.get_filter_changes(filter.filter_id)
+    assert len(all_pending) == 0
```
yihuang, line 81: it should contains the pending transactions?

integration_tests/test_mempool.py
linfeng-crypto: this is what confused me a lot too, if I don't call `wait_for_transaction_receipt`(tx in block mined), `get_new_entries()` returns empty, even if I wait for 15 secs(before the block mined). It seems that the pending filter will not return the pending tx until the tx is mined.

integration_tests/test_mempool.py
linfeng-crypto: I did it before like this, I send tx in a loop, and it runs very quickly, if I get nonce from the API each time, the API will return the same nonce, so that send tx will fail.

integration_tests/test_mempool.py
linfeng-crypto: As I said before, it seems that the filter will not trigger the pending tx event until it was mined. I'm also confused.

integration_tests/test_mempool.py
yihuang: it might be a bug, or instability of the api.

linfeng-crypto:
@yihuang any idea that `pystarport init` error?
https://github.com/crypto-org-chain/cronos/runs/6517985256?check_suite_focus=true
In my personal account, it's ok:
https://github.com/linfeng-crypto/cronos/runs/6504523787?check_suite_focus=true#step:5:2235

integration_tests/configs/long_timeout_commit.yaml
```diff
@@ -0,0 +1,66 @@
...
+            - denom: "basetcro"
+              amount: "1"
+      transfer:
+        params:
+          receive_enabled: true
+          send_enabled: true
+      feemarket:
+        params:
+          no_base_fee: false
+          initial_base_fee: 100000000000
```
yihuang, line 66:
```suggestion
          base_fee: 100000000000
```

integration_tests/configs/long_timeout_commit.yaml
```diff
@@ -0,0 +1,66 @@
...
+            - denom: "basetcro"
+              amount: "1"
+      transfer:
+        params:
+          receive_enabled: true
+          send_enabled: true
+      feemarket:
+        params:
+          no_base_fee: false
+          base_fee: 100000000000
```
yihuang, line 66:
```suggestion
          base_fee: "100000000000"
```

yihuang (approved):

yihuang:
@linfeng-crypto it seems your commits are not signed.

integration_tests/configs/long_timeout_commit.yaml
```diff
@@ -0,0 +1,66 @@
...
```
yihuang, line 66:
```suggestion
          base_fee: "100000000000"
```

PR #503, opened 2022-05-24
Problem: failed tx are ignored in json-rpc apis (backport #502)

yihuang:
Solution:
- fix the issue in ethermint by emitting extra events in ante handler.
- add integration test to test it.
- update cosmos-sdk dependency

👮🏻👮🏻👮🏻 !!!! REFERENCE THE PROBLEM YOUR ARE SOLVING IN THE PR TITLE AND DESCRIBE YOUR SOLUTION HERE !!!! DO NOT FORGET !!!! 👮🏻👮🏻👮🏻

# PR Checklist:

- [ ] Have you read the [CONTRIBUTING.md](https://github.com/crypto-org-chain/chain-main/blob/master/CONTRIBUTING.md)?
- [ ] Does your PR follow the [C4 patch requirements](https://rfc.zeromq.org/spec:42/C4/#23-patch-requirements)?
- [ ] Have you rebased your work on top of the latest master? 
- [ ] Have you checked your code compiles? (`make`)
- [ ] Have you included tests for any non-trivial functionality?
- [ ] Have you checked your code passes the unit tests? (`make test`)
- [ ] Have you checked your code formatting is correct? (`go fmt`)
- [ ] Have you checked your basic code style is fine? (`golangci-lint run`)
- [ ] If you added any dependencies, have you checked they do not contain any known vulnerabilities? (`go list -json -m all | nancy sleuth`)
- [ ] If your changes affect the client infrastructure, have you run the integration test?
- [ ] If your changes affect public APIs, does your PR follow the [C4 evolution of public contracts](https://rfc.zeromq.org/spec:42/C4/#26-evolution-of-public-contracts)?
- [ ] If your code changes public APIs, have you incremented the crate version numbers and documented your changes in the [CHANGELOG.md](https://github.com/crypto-org-chain/chain-main/blob/master/CHANGELOG.md)?
- [ ] If you are contributing for the first time, please read the agreement in [CONTRIBUTING.md](https://github.com/crypto-org-chain/chain-main/blob/master/CONTRIBUTING.md) now and add a comment to this pull request stating that your PR is in accordance with the [Developer's Certificate of Origin](https://github.com/crypto-org-chain/chain-main/blob/master/CONTRIBUTING.md#developer-certificate-of-originn).

Thank you for your code, it's appreciated! :)

PR #513, opened 2022-05-26
Problem: no command to patch missing transactions between v0.7.0 and v0.7.1

yihuang:
Solution:
- add fix-unlucky-tx command to patch txs post v0.7.0 upgrade.

# PR Checklist:


cmd/cronosd/cmd/fix-unlucky-tx.go
```diff
@@ -0,0 +1,209 @@
...
+)
+
+const ExceedBlockGasLimitError = "out of gas in location: block gas meter; gasWanted:"
+
+// FixUnluckyTxCmd update the tx execution result of false-failed tx in tendermint db
+func FixUnluckyTxCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "fix-unlucky-tx [blocks-file]",
+		Short: "Fix tx execution result of false-failed tx after v0.7.0 upgrade, \"-\" means stdin.",
+		Long:  "Fix tx execution result of false-failed tx after v0.7.0 upgrade.\nWARNING: don't use this command to patch blocks generated before v0.7.0 upgrade",
```
thomas-nguy, line 35: as a safe guard, would it be better to hardcode the upgrade height number and skip the command if the height is below this number?

thomas-nguy, line 35: as a safe guard, would it be better to hardcode the upgrade height number and skip the command if the height is below this number?

yihuang: that would complicate the code a little bit though.
To do conditional compile with network build tag, we must create separate modules right?

tomtau: maybe just one for mainnet + have a `--force` flag to skip the safe guard?

thomas-nguy: hmm since we specify the chainId, maybe a simple map chainId -> number would be enough?

thomas-nguy: what would happen if someone try to patch a tx before v0.7.0 upgrade? can it leads to possible state corruption?

yihuang: added a `--min-block-height` as safty guard, with default value of `2693800`.

cmd/cronosd/cmd/fix-unlucky-tx.go
```diff
@@ -0,0 +1,223 @@
...
+							ethTxIndex := txIndex + int64(msgIndex)
+							ethTx, ok := msg.(*evmtypes.MsgEthereumTx)
+							if !ok {
+								continue
+							}
+							evt := abci.Event{
+								Type: evmtypes.TypeMsgEthereumTx,
+								Attributes: []abci.EventAttribute{
+									{Key: []byte(evmtypes.AttributeKeyEthereumTxHash), Value: []byte(ethTx.Hash), Index: true},
+									{Key: []byte(evmtypes.AttributeKeyTxIndex), Value: []byte(strconv.FormatInt(ethTxIndex, 10)), Index: true},
```
yihuang, line 120: This event format emulates the event format the tx is supposed to emit after this PR: https://github.com/tharsis/ethermint/pull/1062

cmd/cronosd/cmd/fix-unlucky-tx.go
```diff
@@ -0,0 +1,223 @@
...
+							Tx:     blk.Txs[i],
+							Result: *txResult,
+						}); err != nil {
+							return err
+						}
+						for _, msg := range tx.GetMsgs() {
+							fmt.Println("patched", blockNumber, msg.(*evmtypes.MsgEthereumTx).Hash)
+						}
+						break
+					} else if txResult.Code != 0 {
```
yihuang, line 140:
```suggestion
					} else if txResult.Code == 0 {
```

integration_tests/test_replay_block.py
```diff
@@ -92,3 +110,20 @@ def test_replay_block(custom_cronos):
...
+    # the second tx is patched
+    assert results[0][1] == txhashes[1].hex()
+    # start the node0 again
+    supervisorctl(
+        custom_cronos.base_dir / "../tasks.ini", "start", "cronos_777-1-node0"
+    )
+    # wait for tm-rpc port
+    wait_for_port(ports.rpc_port(custom_cronos.base_port(0)))
+    # check the tx indexer
+    assert len(cli.txs(f"ethereum_tx.ethereumTxHash={txhashes[1].hex()}")["txs"]) == 1
```
yihuang, line 129: This integration test only works before the root cause is fixed: https://github.com/crypto-org-chain/cronos/pull/503

cmd/cronosd/cmd/fix-unlucky-tx.go
thomas-nguy: from your comment below, I understand that doing it on block before ` v0.7.0` could lead to state corruption because of the event format?

in that case, having the ability to override `--min-block-height` could be dangerous if bad interpreted, unless there is a reason for node operator to use it?

thomas-nguy: maybe at least, strongly enforce the min-block-height when chain-id `cronosmainnet_25-1` is specified

yihuang: we'll provide an external doc to user: https://github.com/crypto-org-chain/cronos/wiki/Patch-Unlucky-Tx#post-v070-upgrade, the users should follow the procedures in the doc.

cmd/cronosd/cmd/fix-unlucky-tx.go
thomas-nguy: one thing I would like to clarify. By adding extra events, we are not modifying the block hash? 

Are the events taken out from the merkle tree computation? (I remember they were removed or added at some point)

yihuang: https://github.com/tendermint/tendermint/blob/044b12585ffde860fa02333b449d5ef7f1e4da7d/types/results.go#L47
yes, only these four fields are considered as deterministic.

cmd/cronosd/cmd/fix-unlucky-tx.go
```diff
@@ -0,0 +1,223 @@
...
+							return err
+						}
+						for _, msg := range tx.GetMsgs() {
+							fmt.Println("patched", blockNumber, msg.(*evmtypes.MsgEthereumTx).Hash)
+						}
+						break
+					} else if txResult.Code == 0 {
+						// find the correct tx index
+						for _, evt := range txResult.Events {
+							if evt.Type == evmtypes.TypeMsgEthereumTx {
```
thomas-nguy, line 143: why do you need to look for the tx index? is it to double check that that it can be found in the event or because it is not ordered?

thomas-nguy: humm it can be different from i?

yihuang: yes, `i` is just the cosmos tx index, but we need the eth tx index, there are many cases makes them different, failed cosmos tx, non-eth cosmos tx, or batch tx.

thomas-nguy (approved):

PR #532, opened 2022-05-31
Problem: not possible to cancel SendToEthereum transactions (fixes #389)

### LINKED ISSUE #389: Problem: not possible to cancel SendToEthereum transactions
Implement a new events in the EVM hooks to be able to cancel an "unbatched" transaction.

Critical points are:
- Verify the sender address and allows to unbatch only transactions related to him.
- Contract needs also to refund the user if it is successfull

thomas-nguy:

Solve this issue https://github.com/crypto-org-chain/cronos/issues/389

# PR Checklist:


x/cronos/types/interfaces.go
```diff
@@ -60,7 +61,8 @@ type EvmLogHandler interface {
 	// Return the id of the log signature it handles
 	EventID() common.Hash
 	// Process the log
-	Handle(ctx sdk.Context, contract common.Address, data []byte, addLogToReceipt func(contractAddress common.Address, logSig common.Hash, logData []byte)) error
+	Handle(ctx sdk.Context, sender common.Address, contract common.Address, data []byte,
```
thomas-nguy, line 64: the reason we are adding the field `sender common.Address` is because in the current ERC20.sol, we are not passing msg.sender to `SendToEthereumEvent` https://github.com/crypto-org-chain/cronos/blob/main/contracts/src/ModuleCRC20.sol#L44

While msg.sender is not always equal to tx's from, this is the compromise to avoid migrating all existing CRC20 contract

scripts/cronos-experimental-devnet.yaml
```diff
@@ -31,6 +31,7 @@ cronos_777-1:
       block:
         max_bytes: "1048576"
         max_gas: "81500000"
+        time_iota_ms: "2000"
```
thomas-nguy, line 34: increase block time to reduce the likehood that tx is immediately included in batch for integration tests and making the cancellation impossible

tomtau (commented):
I read through the issue, and I'm not sure what it entails:
1. Why cancelling SendToEthereum is needed?
2. What are the alternatives? The issue mentions it'd require some architectural changes, but I'm not sure what would need to be changed.
3. What is the tradeoff here? To my limited understanding: e.g. if there are smart contracts SC1 and SC2, SC1 holds tokens, SC2 is used to trigger `SendToEthereum` and then to cancel it (i.e. it's the msg.sender), the tokens will be refunded to SC2 (instead of SC1), is that correct? Can't it be used in some attack scenarios?

If the problem/solution isn't trivial, maybe ADR can be good to clarify the choice?

thomas-nguy:
Sure, let me try to clarify some of your questions

1/ "SendToEthereum" transactions does not have timeout, it assumes it will eventually be included in batch and executed on Ethereum. However it is possible for the transaction to be stuck forever in the transaction pool if the fee specified is too low and never get included in batch. The cancellation can be trigger by the user sending the "SendToEthereum" transaction to be able to delete the tx from the pool and get a refund.

2/ @yihuang mentions letting the user directly send a "CancelSendToEthereum" signing with cosmos wallet. But because the funds are actually managed by the CRC20 contract, not the user, it is not easy to do without the performing the transfer back to the user beforehand, thus using the "evm-hook".  I guess another benefits with "evm-hook" is that user can only use metamask wallet to perform all operations.

3/ The controversial point here is that I am using the field "tx.From" instead of "msg.sender" in the hook because it is not possible to know "msg.sender" with the current event format (it is not specified). 
Since many CRC20 already uses the ModuleCRC20.sol, it is not easy to change without migrating to a new contract (with user consent?)

Consequences: 
It means that if SC1 holds tokens for SC2 and we want to transfer SC2 tokens to some "ethereum address". In case of refund, the one receiving back the CRC20 won't be SC2 but the one that has trigger the transaction.

The solution in that case is not allows to cancel for SC2 (in the guideline, a SC should not expose cancellation function) or have ACL protections.

Open to suggestion here.

tomtau:
> However it is possible for the transaction to be stuck forever in the transaction pool if the fee specified is too low and never get included in batch.

Is that a potential DoS vector?

> The controversial point here is that I am using the field "tx.From" instead of "msg.sender"

ok, that's a potential Pandora's box in terms of security.

> Since many CRC20 already uses the ModuleCRC20.sol, it is not easy to change without migrating to a new contract (with user consent?)

it wouldn't be controversial if there was CRC21 (or however these token interfaces are numbered) that all it did was to patch SendToEthereum-related issues, and the existing (all or just some selected) contracts owned by x/cronos would be auto-migrated to it in a network upgrade

thomas-nguy:
```
Is that a potential DoS vector?
```

If there is no transaction in the pool, then empty or low fee will be included in the next batch. Batch creation do have a limitation in term of speed (100 tx per batch per token every x blocks), that is why it is prioritising the highest fee and letting the user the possibility to cancel their transactions to specify a higher fee.

```
The controversial point here is that I am using the field "tx.From" instead of "msg.sender"

ok, that's a potential Pandora's box in terms of security.
```

After thinking, there is indeed an issue in term of security in case SC2 hold fund in SC1 and try to call "SendToEthereum".
Maybe we should forbid this pattern by adding more comment in the CRC20 specification.


```
it wouldn't be controversial if there was CRC21 (or however these token interfaces are numbered) that all it did was to patch SendToEthereum-related issues, and the existing (all or just some selected) contracts owned by x/cronos would be auto-migrated to it in a network upgrade
```

I think the main issue is that CRC20 does not have migration functions so it wont be easy to migrate fund from CRC20 to a CRC21 without user manually doing it?

Anyways I converted this PR to draft to allow more discussions

yihuang:
> Is that a potential DoS vector?

it does cost some basic gas fee as tx in general and locks up the coins to be sent, although the value of the coins could be near zero.

yihuang:
> 3/ The controversial point here is that I am using the field "tx.From" instead of "msg.sender" in the hook because it is not possible to know "msg.sender" with the current event format (it is not specified).

we should add the `msg.sender` to `__CronosSendToEthereum` event then, it should be alright to change this, since we haven't released the gravity bridge yet.

thomas-nguy:
```
3/ The controversial point here is that I am using the field "tx.From" instead of "msg.sender" in the hook because it is not possible to know "msg.sender" with the current event format (it is not specified).

we should add the msg.sender to __CronosSendToEthereum event then, it should be ok to change, since we haven't released the gravity bridge yet.
```

But we have already few CRC20 contract in production such as WETH or WBTC

yihuang:
> I think the main issue is that CRC20 does not have migration functions so it wont be easy to migrate fund from CRC20 to a CRC21 without user manually doing it?

Directly migrate bytecode of the known mapped CRC20 contracts in upgrade handler?

contracts/src/ModuleCRC21.sol
```diff
@@ -0,0 +1,34 @@
...
+    function unsafe_burn_v2(address addr, uint amount) internal {
+        // Deduct user's balance without approval
+        require(balanceOf[addr] >= amount, "ds-token-insufficient-balance");
+        balanceOf[addr] = sub(balanceOf[addr], amount);
+        totalSupply = sub(totalSupply, amount);
+        emit Burn(addr, amount);
+    }
+
+    // send to ethereum through gravity bridge
+    function send_to_ethereum_v2(address recipient, uint amount, uint bridge_fee) external {
```
thomas-nguy, line 25: Without migration, it is possible to create a wrapper over the CRC20 to be able to call `__CronosSendToEthereum` and have a separate supply from the original CRC20

It would be cleaner to remove `send_to_ethereum` from the CRC20.sol and remove the _v2 here however this might change the bytecode of the contract (even though nobody is using this method now)

Any opinion @tomtau @yihuang

contracts/src/ModuleCRC21.sol
```diff
@@ -0,0 +1,34 @@
...
+    // make unsafe_burn internal
+    function unsafe_burn_v2(address addr, uint amount) internal {
+        // Deduct user's balance without approval
+        require(balanceOf[addr] >= amount, "ds-token-insufficient-balance");
+        balanceOf[addr] = sub(balanceOf[addr], amount);
+        totalSupply = sub(totalSupply, amount);
+        emit Burn(addr, amount);
+    }
+
+    // send to ethereum through gravity bridge
```
thomas-nguy, line 24: Would it be better to make it more generic such as "send_to_chain(....uint chainId)" considering that multi bridge is on the roadmap?

contracts/src/ModuleCRC21.sol
yihuang: Agree with the standard extension, and deprecate the `send_to_ethereum` from the original one.
But existing contracts still need to migrate code to support the gravity bridge, right?

contracts/src/ModuleCRC21.sol
yihuang: sounds good, and we don't have to use the `_v2` suffix.

contracts/src/ModuleCRC21.sol
thomas-nguy: another idea is to have a wrapper over exsting CRC20 contract to be able to add the ability to bridge. 
For example having  `ETH_VVS` will be a wrapper of `VVS.sol` and expose `send_to_ethereum` and `wrapping/unwrapping` functions
The 1-1 supply control will be guarantee by the smart contract

tomtau: I'd suggest documenting these architectural decisions: https://github.com/crypto-org-chain/cronos/blob/main/docs/architecture/adr-template.md + getting some feedback from dApp developers on it

yihuang: > another idea is to have a wrapper over exsting CRC20 contract to be able to add the ability to bridge. For example having `ETH_VVS` will be a wrapper of `VVS.sol` and expose `send_to_ethereum` and `wrapping/unwrapping` functions The 1-1 supply control will be guarantee by the smart contract

it's still a contract migration, the client need to use a different contract address, but internally it reuse the old contract to avoid state migration, right?

thomas-nguy: ```the client need to use a different contract address, but internally it reuse the old contract to avoid state migration```

Why does it need to use old contract? Mapping between gravitydenom <-> contract address will be done on the wrapped contract.
The drawback is that to be able to send to ethereum, one needs to wrap its token first (or unwrap for the opposite direction) but that can be integrated seeminglessly by the Dapp

yihuang: I see, you mean a 1-to-1 token wrapping contract, I guess for erc20, user will need to do a) approve, b) swap, c) send_to_ethereum, right?

integration_tests/test_gravity.py
```diff
@@ -396,8 +394,106 @@ def test_direct_token_mapping(gravity):
...
+        # __CronosSendToChainResponse
+        assert len(txreceipt.logs) == 3
+        tx_id = get_id_from_receipt(txreceipt)
+        assert txreceipt.status == 1, "should success"
+
+        def check_deduction():
+            v = crc21_contract.caller.balanceOf(ADDRS["community"])
+            return v == 0
+
+        wait_for_fn("check deduction", check_deduction)
```
yihuang, line 485: should be able to check balance immediately, no need to wait?

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 - [cronos#418](https://github.com/crypto-org-chain/cronos/pull/418) Support logs in evm-hooks and return id for SendToEthereum events
 - [cronos#489](https://github.com/crypto-org-chain/cronos/pull/489) Enable jemalloc memory allocator, and update rocksdb src to `v6.29.5`.
 - [cronos#511](https://github.com/crypto-org-chain/cronos/pull/511) Replace ibc-hook with ibc middleware, use ibc-go upstream version.
+- [cronos#532](https://github.com/crypto-org-chain/cronos/pull/532) Add SendtoChain and CancelSendToChain support from evm call.
```
yihuang, line 12: I think it's "State Machine Breaking", also the above one.

scripts/cronos-experimental-devnet.yaml
yihuang:
```suggestion
        # increase block time to reduce the likehood that tx is immediately included in batch for integration tests and making the cancellation impossible
        time_iota_ms: "2000"
```

x/cronos/keeper/evm.go
```diff
@@ -54,13 +54,13 @@ func (k Keeper) CallModuleCRC20(ctx sdk.Context, contract common.Address, method
 	return res.Ret, nil
 }
 
-// DeployModuleCRC20 deploy an embed erc20 contract
-func (k Keeper) DeployModuleCRC20(ctx sdk.Context, denom string) (common.Address, error) {
-	ctor, err := types.ModuleCRC20Contract.ABI.Pack("", denom, uint8(0))
+// DeployModuleCRC21 deploy an embed erc20 contract
```
yihuang, line 57:
```suggestion
// DeployModuleCRC21 deploy an embed crc21 contract
```

x/cronos/types/contracts/ModuleCRC20.json
```diff
@@ -687,5 +687,5 @@
...
...fffffffffffffffffff168152602001908152602001600020541015612927576040517f08c379a000000000000000000000000000000000000000000000000000000000815260040180806020018281038252601d8152602001807f64732d746f6b656e2d696e73756666696369656e742d62616c616e636500000081525060200191505060405180910390fd5b612970600360008473ffffffffffffffffffffffffffffffffffffffff1673ffffffffffffffffffffffffffffffffffffffff168152602001908152602001600020548261276c565b600360008473ffffffffffffffffffffffffffffffffffffffff1673ffffffffffffffffffffffffffffffffffffffff168152602001908152602001600020819055506129bf6002548261276c565b6002819055508173ffffffffffffffffffffffffffffffffffffffff167fcc16f5dbb4873280815c1ee09dbd06736cffcc184412cf7a71a0fdb75d397ca5826040518082815260200191505060405180910390a25050565b828054600181600116156101000203166002900490600052602060002090601f016020900481019282601f10612a5857805160ff1916838001178555612a86565b82800160010185558215612a86579182015b82811115612a85578251825591602001919060010190612a6a565b5b509050612a939190612a97565b5090565b612ab991905b80821115612ab5576000816000905550600101612a9d565b5090565b9056fea2646970667358221220b0bcda9e8fad608c29f91994591f62951230c4ad72859af8c34245277302a33f64736f6c63430006080033"
```
yihuang, line 690: https://docs.soliditylang.org/en/develop/metadata.html
the reason that code comments change leads to bytecode change/

yihuang (changes requested):
LGTM in general, only some minor suggestions.

integration_tests/test_gravity.py
thomas-nguy: yes, changed 👌

CHANGELOG.md
thomas-nguy: moved to state breaking change

x/cronos/types/contracts/ModuleCRC20.json
thomas-nguy: I see, okay let me remove the comment for now

thomas-nguy:
[client/docs/statik/statik.go](https://github.com/crypto-org-chain/cronos/pull/532/files#diff-fa3f9d3fbea063fbc470f21a8e6fb36510183cb29a464a46f8a4d4fff38673ca) 

this is modified after running `ofmt -s -w .`

yihuang (approved):
LGTM

PR #533, opened 2022-05-31
Problem: missing ADR draft for a transition to CCV (fixes #194)

### LINKED ISSUE #194: Problem: missing ADR draft for a transition to CCV
Cronos to be a child chain.

https://github.com/cosmos/interchain-security
https://github.com/cosmos/ibc-go/pull/116 
https://github.com/informalsystems/cross-chain-validation/pull/26/files

tomtau:
Solution: sketched out a draft for CCV usage where Cronos is the provider chain
for Gravity Bridge.
(note that the original issue considered Cronos could be a consumer chain
which may still be a valid option, but the complexity of transition may be too high.)

### DIFF
--- a/docs/architecture/adr-005.md
+++ b/docs/architecture/adr-005.md
@@ -48,6 +48,7 @@ Proposed
 
 - Complexity and maintenance overhead of a separate network codebase and operation. (Note that the Gravity Bridge infrastructure demands an inherent operational complexity, e.g. the need to run Ethereum full nodes, regardless of where the module is included.)
 - A slight increase in the cross-chain transfer latency. (The overall time will still be dominated by the operations on Ethereum though.)
+- More complex refund or cancellation workflows, as token transfers can get stuck in IBC or Ethereum submissions.
 
 ### Neutral
 
--- a/docs/architecture/adr-005.md
+++ b/docs/architecture/adr-005.md
@@ -1,0 +1,61 @@
+# ADR 005: Cross-chain Validation for Gravity Bridge
+
+## Changelog
+* 2022-05-31: first draft
+
+## Context
+
+Cross-chain Validation is a shared security mechanism over IBC that leads to the creation of "provider" and "consumer" blockchains.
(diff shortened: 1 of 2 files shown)


docs/architecture/adr-005.md
```diff
@@ -0,0 +1,61 @@
...
+
+- Isolation of Gravity Bridge: any potential issues discovered in it will not affect the overall Cronos network's operation.
+- Independent processing of transactions: any transaction spikes in the EVM or Gravity Bridge modules will not affect each other directly.
+- Storage sharding: the storage needed by Gravity Bridge can live on a different node from the one executing Cronos with the EVM module.
+- Simplified interface for cross-chain token transfers: only IBC-related logic needs to be considered inside EVM.
+- Independent upgrades: any breaking changes in Gravity Bridge will not demand upgrades on Cronos (and vice versa).
+
+### Negative
+
+- Complexity and maintenance overhead of a separate network codebase and operation. (Note that the Gravity Bridge infrastructure demands an inherent operational complexity, e.g. the need to run Ethereum full nodes, regardless of where the module is included.)
```
thomas-nguy, line 49: I think that summarise the pro and cons. 

While technically all "macro" features in Cronos could be created into a separate chain. We need to evaluate if there the gravity bridge as a standalone chain can have its own roadmap (future enhancement, new features etc) and thus could justify the operation costs (server, maintenance, monitoring, data storage cost)

tomtau: https://github.com/Gravity-Bridge/Gravity-Bridge/issues?q=is%3Aissue+is%3Aopen+label%3Aenhancement

I assume one future breaking change will be ERC721 and ERC1155 transfers?

thomas-nguy: + multi chain support maybe

tomtau:
updated the status to rejected -- based on the Slack discussion, this would impact the implementation timeline, so the preference is to keep the current module (as in [ADR 001](https://github.com/crypto-org-chain/cronos/blob/main/docs/architecture/adr-001.md)).
@thomas-nguy is it ok to merge it?

thomas-nguy (commented):
lgtm

thomas-nguy (approved):

PR #545, opened 2022-06-09
Problem: no decision record for generic event format for evm-hook

thomas-nguy:

# PR Checklist:

### DIFF
--- a/docs/architecture/adr-007.md
+++ b/docs/architecture/adr-007.md
@@ -8,7 +8,7 @@
 The current evm-hooks relies on event emitted from a smart contract to trigger specific actions upon contract execution.
 The current way to define those events is not flexible: events are hardcoded, parameters fixed and correspond only to one specific action.
 
-This can lead to potential issue in the long run when commands needs to be updated to add more parameters (ibc-go/ gravity bridge) or even if we need to implement more "actions". Since contracts are immutable, contract developers won't be able to have access to these functions after deploying their CRC20 contracts.
+This can lead to potential issue in the long run when commands needs to be updated to add more parameters (ibc-go/ gravity bridge) or even if we need to implement more "actions". Since contracts are immutable, contract developers won't be able to have access to these functionalities after deploying their CRC20 contracts if we perform an upgrade.
 
 In this ADR, we define generic events format that can be use by smart contract developer to call any "action" upon successful contract execution. 
 
--- a/docs/architecture/adr-007.md
+++ b/docs/architecture/adr-007.md
@@ -28,7 +28,7 @@ __CronosCallAfterPayment(string functionName, address payer, uint amount, byte[]
 - **functionName** corresponds to the equivalent hardcoded eventName such as `CronosSendToIBC` for the event `__CronosSendToIbc` or `CronosSendToChain` for the event `__CronosSendToChain`
 - **payer** is the address which make the payment, usually equal to `msg.sender`
 - **amount** corresponds to the `amount` that the payer has paid (either by burning or sending asset to the SC)
-- **args** corresponds to the argument in order of the event removing the `sender` and `amount` from the parameters
+- **args** corresponds to the argument in order of the event removing the `sender` and `amount` from the parameters encoded using the abi event pack
 
 
 
--- a/docs/architecture/adr-007.md
+++ b/docs/architecture/adr-007.md
@@ -53,9 +53,10 @@ Contract Demo is DSToken {
  }
  
  // send an "amount" of the contract token to recipient through IBC
- function send_to_ibc(byte[] recipient) public {
+ // args should contains the encoded value recipient or extra value if needed in the future
+ function send_to_ibc(byte[] args) public {
      unsafe_burn(msg.sender, amount);
-     emit __CronosCallAfterPayment("CronosSendToIbc", msg.sender, amount, recipient);
+     emit __CronosCallAfterPayment("CronosSendToIbc", msg.sender, amount, args);
  }
 
 ```
--- a/docs/architecture/adr-007.md
+++ b/docs/architecture/adr-007.md
@@ -54,13 +54,14 @@ Contract Demo is DSToken {
  
  // send an "amount" of the contract token to recipient through IBC
  // args should contains the encoded value recipient or extra value if needed in the future
- function send_to_ibc(byte[] args) public {
+ function send_to_ibc(uint amount, byte[] args) public {
      unsafe_burn(msg.sender, amount);
      emit __CronosCallAfterPayment("CronosSendToIbc", msg.sender, amount, args);
  }
 
 ```
 
+In this example byte[] args should contains the encoded recipient address from `event __CronosSendToIbc(address sender, string recipient, uint256 amount)` but in the future, if we needs to add extra parameters such as `fee_amount` or `fee_recipient` we will be able to use the same method in the smart contract, but encode those extra parameters in args. 
 
 On the chain side, the events parameters are decoded and uses to call a specific actions. 
 
--- a/docs/architecture/adr-007.md
+++ b/docs/architecture/adr-007.md
@@ -18,10 +18,10 @@ We define 2 specifics events that smart contract can use to trigger specific act
 
 ```
 # Trigger an action after contract execution which only involve the sender account
-__CronosCall(string functionName, address sender,  byte[]args)
+__CosmosCall(string functionName, address sender,  byte[]args)
 
 # Trigger an action after contract execution that require moving asset from the smart contract "account"
-__CronosCallAfterPayment(string functionName, address payer, uint amount, byte[]args)
+__CosmosCallAfterPayment(string functionName, address payer, uint amount, byte[]args)
 
 ```
 
@@ -41,7 +41,7 @@ import "ds-token/token.sol";
 
 Contract Demo is DSToken {
 
- __CronosCallAfterPayment(string functionName, address sender, uint amount, byte[]args)
+ __CosmosCallAfterPayment(string functionName, address sender, uint amount, byte[]args)
  
  // unsafe_burn burn tokens without user's approval and authentication, used internally
  function unsafe_burn(address addr, uint amount) internal {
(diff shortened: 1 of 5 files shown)


docs/architecture/adr-007.md
```diff
@@ -0,0 +1,93 @@
...
+     require(balanceOf[addr] >= amount, "ds-token-insufficient-balance");
+     balanceOf[addr] = sub(balanceOf[addr], amount);
+     totalSupply = sub(totalSupply, amount);
+     emit Burn(addr, amount);
+ }
+ 
+ // send an "amount" of the contract token to recipient through IBC
+ function send_to_ibc(byte[] recipient) public {
+     unsafe_burn(msg.sender, amount);
+     emit __CronosCallAfterPayment("CronosSendToIbc", msg.sender, amount, recipient);
```
yihuang, line 58: This example is no difference then before, what this proposal really enables is the smart contract can provide a generic API like `call_cronos` where end user can supply the `functionName` and generic args, right?

thomas-nguy: I changed the field name to be more generic

yihuang: I think what we really do here is to solve the upgradeability issue, which is to avoid upgrading dapp's contract when the chain added new capabilities to these apis, right?
So I suggest putting more emphasis on the upgradeability issue, and demonstratee in the example.

docs/architecture/adr-007.md
```diff
@@ -0,0 +1,96 @@
...
+     balanceOf[addr] = sub(balanceOf[addr], amount);
+     totalSupply = sub(totalSupply, amount);
+     emit Burn(addr, amount);
+ }
+ 
+ // send an "amount" of the contract token to recipient through IBC
+ // args should contains the encoded value recipient or extra value if needed in the future
+ function send_to_ibc(uint amount, byte[] args) public {
+     unsafe_burn(msg.sender, amount);
+     emit __CronosCallAfterPayment("CronosSendToIbc", msg.sender, amount, args);
```
yihuang, line 59: Will we recommend apps to allow user to pass in the function name as well?

thomas-nguy: I guess it depend of the smart contract intent.

but the function name can also be an input parameter if the smart contract is a controller for ica for example and they expects more features to be implemented

yihuang: It's about convention and expectation, for example, we want to extend sth for ica, would we add a new function name, or just support extend the args format, in the first case, this demo contract still need to upgrade itself.

thomas-nguy: yeah I guess it also depend on what kind of actions we need to implement in the future

I guess for now the focus is on "native token transfer capability" from smart contract account to a "recipient".

If we support both way (hardcoded event or generic events), the sc developers can either choose between exposing limited functions and define different payment rule for each transfer methods (but it takes the risk to not be compatible in the future) or have a more generic capability that does not requires contract upgrade

yihuang: what if we just add a some really generic stuff like this:

```
event __CronosGenericCall(address, uint, byte[])

function cronos_generic_call(uint amount, byte[] payload) public {
  if amount > 0 {
    unsafe_burn(msg.sender, amount);
  }
  emit __CronosGenericCall(msg.sender, amount, payload);
}
```

thomas-nguy: you mean putting the function name in byte[]?

I guess this is fine, we need a better way to serialised the function name and input parameters available for smart contract devs

yihuang: and don't need to separate `__CronosCall` and `__CronosCallAfterPayment`.

thomas-nguy: mmm the reason it is separated is for clarity. It is not clear what is the "amount" if you call name the function `__CronosCall` 

I also expect that having a generic function `__CronosCall` is not impossible at our stage and we might need to implement more "generic events" in the future for different type of actions

yihuang: the sematic is similar to calling a smart contract method with a value.

yihuang: anyway, it's a trade-off here, too generic may mean unexpected behavior in the future, and less generic means smart contract upgrade, personally, I prefer explicit contract upgrade to implicit stuff actually.
when an app wants to use a non-trivial new feature like gravity-bridge, it's reasonable to ask for an upgrade.
It's just the chain has gov module already, so we can do the upgrade relatively painlessly. but the smart contract can do govenance too, just ask the community to vote or sth.

thomas-nguy:
```
But the smart contract can do govenance too, allowing the community to vote or sth.
```

Do you mean bytecode replacement? if its case by case it seems hard to do... and also controversial

I guess our problem reside on the CRC20 contract that are not supposed to be upgradable.

To summarise from the less generic to the most generic solution:

1. Just make the current events "more flexible" for example `__CronosSendToIbc(address, uint, bytes[]arg)` to make sure they will still be compatible after upgrade

2. Make a generic function only for token related stuff ` __CronosCallAfterPayment(string functionName, address sender, uint amount, byte[]args)` which can be used to perform actions on contract's assets which is basically what this ADR propose

3. Try to define a unique generic event (at least best effort) that can be use by contract developers such as `__CronosGenericCall(address, uint, byte[])` or `__CronosGenericCall(byte[], byte[])` which allows to call any methods and future hooks from the smart contract

yihuang: > Do you mean bytecode replacement?

I mean the smart contract should do the upgrade for themselves if they want to bring in a significant new feature like gravity-bridge.
What we are proposing here is trying to smuggle new features into the smart contract through chain upgrade.

thomas-nguy:
```
What we are discussing here is trying to smuggle new features into the smart contract through chain upgrade.
```

Technically those event hooks are a way for user to chain a evm tx with a cosmos tx upon contract execution

I can see two big use cases

- Trigger an action on behalf msg.sender account. In that case, using the smart contract as a gateway is just a way for us to "expose" those functionalities to metamask (in the effort to be compatible with ethereum)

- Trigger an action on behalf of the smart contract account. Smart contracts have the ability to move assets (coin, nft etc) from their accounts and perform diverse actions (ibc, bridge etc)

In that case does is it okay to give those abilities to all the smart contracts in general? the only issue is that for new features, previous deployed smart contracts won't have access to them. It is maybe okay if we consider that those features was not in the scope of the Dapps developers at the first place but it becomes a problem when we modify existing features (with the need to add or remove parameters for example) or when those features are more "user" centrics and we deprives old contract from new "cool" features that we implement.

But this is a debatable topic.

docs/architecture/adr-007.md
```diff
@@ -0,0 +1,96 @@
...
+
+On the chain side, the events parameters are decoded and uses to call a specific actions. 
+
+List of possible action (as of today)
+
+- Actions : `CronosCancelSendToChain`
+- Actions after Payment: `CronosSendToAccount`, `CronosSendToIbc`, `CronosSendCroToIbc`, `CronosSendToChain`
+
+ 
+With this generic event format, it is possible to add and remove the parameters for those action in the future or even add new actions while "upgrading" the chain. Smart contract won't be force to migrate to a new contract to benefit those functionnalities.
```
calvinaco, line 74: I am trying to understand how the generic can help support future event format without the need of upgrading the contract.

Contract should always validate the arguments being passed before they emit the event, right?

thomas-nguy: no only a subset of the arguments needs to be validated. If we consider that the action is done from msg.sender account (after receiving or not cosmos asset). then most likely those fields will be specify by the users and won't need validation.

calvinaco: On a second read, I think this ADR is not intended to avoid the need of contract upgrade when event signatures is changed on upgrade, am I right?

It's just providing a generic way to emit event?

thomas-nguy: this ADR suggests a more flexible to define events for evm-hooks, and one of the issue indeed is that we cannot change event signature if we perform an upgrade and some extra parameters needs to be specified by users

calvinaco: > If we consider that the action is done from msg.sender account (after receiving or not cosmos asset). then most likely those fields won't need validation.

This is true for some "chain utility" contract such as the [CroBridge](https://github.com/crypto-org-chain/cronos/blob/main/integration_tests/contracts/contracts/CroBridge.sol). 

But for most other contracts with their own functionality, I think they should safeguard the emitted events and cannot trust user inputs. And the generic event alone could not be sufficient for them. They still need to upgrade their contract to emit the new event upon chain upgrade.

thomas-nguy: yeah this is not possible to define a single generic event that can be use for every case, and sometime upgrading the smart contract is the best way to do it. 

This ADR does not intent to solve all solution but propose a better standard at least for the token transfer functionality which involve amount deduction (burn or payment) from msg.sender and then perform an action (on the cosmos transferred amount from contract address to msg.sender address)

The action is defined by the user so contract only needs to validate the payment (amount + sender address) and user choose what action to do after

calvinaco: On this angle I love the idea of generic event.

The generic event also have the added benefit of making Cosmos EVM hook from smart contract more readable --- If you see the `__CronosCall(...)` you know it is triggering an action on Cosmos layer, instead of seeing the `__CronosSendToIbc(...)` and then the reader needs to dig into different areas for what's impling. Especially in the future when there are more events need to be added.

Extending from this perspective, I am thinking if we want to define the event name to reveal this intention more explicitly, say `__InternalCall()`, `__CosmosCall()` or `__CallCosmos()` etc.

tomtau (commented):
I guess this PR can be closed until we try and know more about the stateful precompile approach that could be an alternative to the hooks?

thomas-nguy:
Didnt we agree to go along with this ADR first and have the precompile as an improvement?

thomas-nguy:
sounds good

yihuang:
https://github.com/crypto-org-chain/cronos/runs/6911115133?check_suite_focus=true

https://eth.wiki/json-rpc/API becomes 404, it seems new version is this one: https://ethereum.github.io/execution-apis/api-documentation/

docs/architecture/adr-007.md
thomas-nguy: sounds good, changed the name to `_CosmosCall`

PR #544, opened 2022-06-08
Problem: no decision record for crc21 standard and crc20 token migration

thomas-nguy:
👮🏻👮🏻👮🏻 !!!! REFERENCE THE PROBLEM YOUR ARE SOLVING IN THE PR TITLE AND DESCRIBE YOUR SOLUTION HERE !!!! DO NOT FORGET !!!! 👮🏻👮🏻👮🏻

Add ADR-006

# PR Checklist:

- [ ] Have you read the [CONTRIBUTING.md](https://github.com/crypto-org-chain/chain-main/blob/master/CONTRIBUTING.md)?
- [ ] Does your PR follow the [C4 patch requirements](https://rfc.zeromq.org/spec:42/C4/#23-patch-requirements)?
- [ ] Have you rebased your work on top of the latest master? 
- [ ] Have you checked your code compiles? (`make`)
- [ ] Have you included tests for any non-trivial functionality?
- [ ] Have you checked your code passes the unit tests? (`make test`)
- [ ] Have you checked your code formatting is correct? (`go fmt`)
- [ ] Have you checked your basic code style is fine? (`golangci-lint run`)
- [ ] If you added any dependencies, have you checked they do not contain any known vulnerabilities? (`go list -json -m all | nancy sleuth`)
- [ ] If your changes affect the client infrastructure, have you run the integration test?
- [ ] If your changes affect public APIs, does your PR follow the [C4 evolution of public contracts](https://rfc.zeromq.org/spec:42/C4/#26-evolution-of-public-contracts)?
- [ ] If your code changes public APIs, have you incremented the crate version numbers and documented your changes in the [CHANGELOG.md](https://github.com/crypto-org-chain/chain-main/blob/master/CHANGELOG.md)?
- [ ] If you are contributing for the first time, please read the agreement in [CONTRIBUTING.md](https://github.com/crypto-org-chain/chain-main/blob/master/CONTRIBUTING.md) now and add a comment to this pull request stating that your PR is in accordance with the [Developer's Certificate of Origin](https://github.com/crypto-org-chain/chain-main/blob/master/CONTRIBUTING.md#developer-certificate-of-originn).

Thank you for your code, it's appreciated! :)

PR #557, opened 2022-06-17
Problem: No decision record for denom and contract mapping for bi-directional token conversion

calvinaco:
Solution: Add ADR 008 to propose the enhancement

# PR Checklist:


docs/architecture/adr-008.md
```diff
@@ -0,0 +1,191 @@
...
+key -> value
+{prefixDenomToExternalContract}{denom} -> {sourceKey};{contract_address}
+{prefixContractToDenom}{contract_address} -> {sourceKey};{denom}
+```
+where `{sourceKey}` is a list of token sources denoted by integer constants:
+```go
+const (
+    sourceKeyCronos = iota + 1  // 1
+    sourceKeyIBC                // 2
+    sourceKeyGravityBridge      // 3
```
yihuang, line 59: ibc and gravity bridge type tokens are distinguishable by prefix of denom: `ibc/***` and `gravity0x***`, I suggest we just define a denom name prefix for the local EVM tokens, `"crc20/" + contract address`.

docs/architecture/adr-008.md
```diff
@@ -0,0 +1,191 @@
...
+
+## Terminologies
+
+### EVM Token
+
+EVM Tokens are tokens that adopt the widely used CRC20/CRC21 token standard. These token are smart contracts deployed on Cronos EVM.
+
+### Local EVM Token
+
+Local EVM Tokens are EVM tokens that are originated rom Cronos EVM, that means the minting of the tokens happen on Cronos.
```
JayT106, line 20: rom -> from

docs/architecture/adr-008.md
```diff
@@ -0,0 +1,191 @@
...
+
+Native Tokens are tokens on the Cosmos layer of Cronos. They are not directly usable on Cronos EVM but an intermediate token format between bridge protocols and Cronos EVM because most bridge protocols are Cosmos module that interacts with native tokens.
+
+## Decision
+
+### 1. Denom and Contract Mapping Key-Value
+
+#### A. Current Design
+
+Currently, the native token denom to contract address is registered in the Cronos module kvstore and stored as key-value paris:
```
JayT106, line 32: paris -> pairs

docs/architecture/adr-008.md
calvinaco: Ok, I think's thats a good point. Then we would require all future tokens to define a unique prefix at their native token, which should be manageable.

calvinaco:
> Overall design looks fine for our specific use cases however there are some sensitive point that need to be discussed:
> 
> While for "non-source" token, it is okay to mint and burn CRC20 tokens upon a transfer because they are essentially voucher, it becomes more problematic when the token is the "source".
> 
> It means that the total supply of the CRC20 token will decrease upon transfer. MarketCap of a token is calculated totalSupply*price of the token but if we start to burn the token during a transfer, we are messing with this metric because the "burned" token still exist, but under another form (cosmos token)
> 
> To have the real market cap of the token, we need to add totalSupply + total cosmos tokens which may be harder to compute.
> 
> Also with this design, it is not possible to have a "control" supply on what is created and what is supposed to be locked.

Thanks for the feedback.

To keep the total supply on EVM level, instead of burning the CRC20/CRC21, we can have new bridge methods and event for it. So the contract need to emit event `send_native_to_ethereum(address recipient, uint amount, uint bridge_fee)` instead and the EVM hook will
1. Transfer the CRC20/CRC21 to the EVM address of the Cronos module account
2. Do the rest of the steps, mint native token, transfer to the user Cosmos address, send to the bridge module etc

This way the total supply  on smart contract level is not affected, and the locked amount can be easily checked by taking a look at the token owned by Cronos module account.

The caveat is that we need extra methods and events defined in CRC20/CRC21 for it, which should be still ok as this is an opt-in feature, similar to the case of https://github.com/crypto-org-chain/cronos/pull/544

> it is not possible to have a "control" supply on what is created and what is supposed to be locked.

May I know what kind of "control" are we referring here? Thanks.

docs/architecture/adr-008.md
calvinaco: Thanks! Resolved

docs/architecture/adr-008.md
calvinaco: Thanks! Resolved

thomas-nguy:
`May I know what kind of "control" are we referring here? Thanks.` 

Could be an invariant in golang, but typically a control check that nothing has been created from thin air (ie total supply cosmos token = total burned CRC20 token)


I think instead of triggering the logic from the evm hook, the easiest way is to change the CRC20 contract 

``` 
    // send to another chain through gravity bridge
    function send_to_chain(address recipient, uint amount, uint bridge_fee, uint chain_id) external {
        unsafe_burn_internal(msg.sender, add(amount, bridge_fee));
        emit __CronosSendToChain(msg.sender, recipient, amount, bridge_fee, chain_id);
    }
 ```
 
 to 
 
 ```
     // send to another chain through gravity bridge
    function send_to_chain(address recipient, uint amount, uint bridge_fee, uint chain_id) external {
        unsafe_transfer(msg.sender, CronosModuleAddress, add(amount, bridge_fee))
        emit __CronosSendToChain(msg.sender, recipient, amount, bridge_fee, chain_id);
    }
 ```
 
 or
 
 ```
        // send to another chain through gravity bridge
    function send_to_chain(address recipient, uint amount, uint bridge_fee, uint chain_id) external {
        unsafe_transfer(msg.sender, CronosModuleAddress, add(amount, bridge_fee))
        emit __CosmosCall("SendToChain", msg.sender, amount,args[recipient,bridge_fee,chain_id]);
    }
```        
 
It uses the same assumption originally that the contract is written correctly and worse case we could use an invariant to check the CronosModuleAddress has the correct supply.

docs/architecture/adr-008.md
yihuang: And maybe we don't need the mapping at all, if we enforce the naming convention `crc20/contract address`.

CeruleanAtMonaco:
For Local EVM Token on Cronos bridged to Ethereum, which ADR captures how the Ethereum side token contract standard should be?

docs/architecture/adr-008.md
```diff
@@ -0,0 +1,191 @@
...
+
+EVM hooks for bridge applications are to be updated accordingly. Upon receiving the emitted bridge events, the EVM hook should:
+1. Unpack and validate the EVM events;
+2. Call `keeper.GetDenomByContract()` to validate and get the token source and native token denom of the EVM token smart contract address;
+3. (Optional) Validate if the native token denom is supported by the bridge^1;
+4. If the token is local EVM token, mint the corresponding native token and send to the user Cosmos address; Otherwise, send the native token from the contract address module account to the user Cosmos address;
+5. Interact with the bridge module and send the token to external chains;
+6. Emit Cosmos events as needed.
+
+^1 It is up to the design of each EVM hook to validate and restrict which the native denom can be bridged with the protocol. They can decide not to put any restriction to allow tokens transferring freely between differnt chains.
```
CeruleanAtMonaco, line 109: typo: differnt -> different

thomas-nguy:
```
For Local EVM Token on Cronos bridged to Ethereum, which ADR captures how the Ethereum side token contract standard should be?
```

There is no standard for Ethereum side token contract. They are direct implementation deployed and controlled by the bridge as CosmosToken.sol (https://github.com/crypto-org-chain/gravity-bridge/blob/v2.0.0-cronos/solidity/contracts/CosmosToken.sol)

IE deploy function

https://github.com/crypto-org-chain/gravity-bridge/blob/v2.0.0-cronos/solidity/contracts/Gravity.sol#L607

Means that developer have no flexibility there

calvinaco:
> `May I know what kind of "control" are we referring here? Thanks.`
> 
> Could be an invariant in golang, but typically a control check that nothing has been created from thin air (ie total supply cosmos token = total burned CRC20 token)
> 
> I think instead of triggering the logic from the evm hook, the easiest way is to change the CRC20 contract
> 
> ```
>     // send to another chain through gravity bridge
>     function send_to_chain(address recipient, uint amount, uint bridge_fee, uint chain_id) external {
>         unsafe_burn_internal(msg.sender, add(amount, bridge_fee));
>         emit __CronosSendToChain(msg.sender, recipient, amount, bridge_fee, chain_id);
>     }
> ```
> 
> to
> 
> ```
>   function source() bool {
>      return _isSource
>   }
> 
>     // send to another chain through gravity bridge
>    function send_to_chain(address recipient, uint amount, uint bridge_fee, uint chain_id) external {
>       if (source()) {
>             unsafe_transfer(msg.sender, CronosModuleAddress, add(amount, bridge_fee))
>        } else {
>             unsafe_burn_internal(msg.sender, add(amount, bridge_fee));
>        }
>        emit __CronosSendToChain(msg.sender, recipient, amount, bridge_fee, chain_id);
>    }
> ```
> 
> or
> 
> ```
>    function source() bool {
>      return _isSource
>   }
>   
>        // send to another chain through gravity bridge
>    function send_to_chain(address recipient, uint amount, uint bridge_fee, uint chain_id) external {
>       if (source()) {
>           unsafe_transfer(msg.sender, CronosModuleAddress, add(amount, bridge_fee))
>        } else {
>           unsafe_burn_internal(msg.sender, add(amount, bridge_fee));
>        }
>        emit __CosmosCall("SendToChain", msg.sender, amount,args[recipient,bridge_fee,chain_id]);
>    }
> ```
> 
> It uses the same assumption originally that the contract is written correctly (We probably need to modify the CRC standard though) and worse case we could use an invariant to check the CronosModuleAddress has the correct supply.

Got your point. Do you have anything in mind how the control could be implemented?

> It uses the same assumption originally that the contract is written correctly

I am thinking for a different scenarios and how the native code can check the correctness of it. But I think the smart contract can still workaround those checks. e.g. The smart contract can have a `burn` method that does nothing; A `totalSupply` query method that returns arbitrary value instead of real total supply.

So at the end instead of relying on the contract to call mint/burn/transfer, if we have the EVM hooks or IBC middlewares to call the contract methods accordingly, we are hoping at least those supply control are in place. And at the end the assumption still needs to hold.

But I am ok with doing the calls on smart contract as well. Because at the end of day probably not much we can do unless we do a bytecode check? to make sure those methods are implemented as expected.

docs/architecture/adr-008.md
```diff
@@ -0,0 +1,191 @@
...
+#### GetDenomByContract
+
+```go
+GetDenomByContract(ctx sdk.Context, contract common.Address)
+    (source int, denom string, found bool)
+```
+
+`source` is added in the returned value to denote the source of the denom to smart contract address mapping
+
+##### GetDenomByContract
```
brianatcrypto, line 142: Is this typo? should be GetContractByDenom?

docs/architecture/adr-008.md
calvinaco: Thanks, updated!

docs/architecture/adr-008.md
calvinaco: Thanks, updated!

docs/architecture/adr-008.md
```diff
@@ -0,0 +1,191 @@
...
+### Negative
+
+- No flexible control of token conversion. For example, we cannot restrict certain token cannot be bridged over particular protocol. It relies on the bridge module to manage the restriction list.
+- Potentially complicated chain upgrade logic.
+
+### Neutral
+
+- Multi-sourced token is not supported. This is a deliberate design decision to avoid complicated supply management logic.
+
+## References
```
tomtau, line 191: maybe one reference to look at: https://github.com/evmos/evmos/tree/main/x/erc20/spec

thomas-nguy:
Should we merge this ADR?

docs/architecture/adr-008.md
```diff
@@ -0,0 +1,191 @@
...
+where `{sourceKey}` is a list of token sources denoted by integer constants:
+```go
+const (
+    sourceKeyCronos = iota + 1  // 1
+    sourceKeyIBC                // 2
+    sourceKeyGravityBridge      // 3
+)
+```
+
+#### C. Design Explain
```
tomtau, line 63:
```suggestion
#### C. Design Explained
```

or 

```suggestion
#### C. Design Explaination
```

docs/architecture/adr-008.md
```diff
@@ -0,0 +1,191 @@
...
+```go
+GetContractByDenom(ctx sdk.Context, denom string)
+    (source int, contract common.Address, found bool)
+```
+
+`source` is added in the returned value to denote the source of the denom to smart contract address mapping
+
+##### ConvertCoinFromNativeToCRC20
+
+In the current design, the method permits only the denom is one of the IBC and Gravity Bridge denoms before doing any conversion. In this ADR, we want to loosen this validation check. If such a requirement exists, it should be performed in the IBC hooks.
```
tomtau, line 153: should this be renamed if the method isn't strictly converting to CRC20 and may be converting to ERC20?

tomtau (commented):
I guess the token conversion should only be for fungible tokens, or should it also consider the new native NFT module: https://github.com/cosmos/cosmos-sdk/tree/main/x/nft/spec ?

yihuang (approved):
LGTM.
One thing to add there's sth similar in the precompiled contract example: https://github.com/yihuang/cronos/blob/test-precompiles/integration_tests/test_precompiles.py

docs/architecture/adr-008.md
```diff
@@ -0,0 +1,173 @@
...
+```go
+BurnVouchersToEvmCoins(ctx sdk.Context, receiver string, coins sdk.Coins) error
+```
+
+This method works for local EVM tokens. It burns the previously-minted native token from the `receiver` address. This method will then invoke the smart contract to mint the EVM tokens to the EVM address of the `receiver`.
+
+The working mechanism is as follows:
+1. Call `keeper.GetContractByDenom()` to validate and get the token source and smart contract address of the provided native token denom;
+2. Burn the voucher native token from the `receiver` Cosmos address;
+3. Invoke the corresponding smart contract `mint_by_cronos_module` method to mint the EVM token to `receiver` EVM address.
```
thomas-nguy, line 149: Before moving to Accepted, need to change this section, the CRC20 token is not burned but moved to a vault.

So we need to move back to recipient address

Issue #553, opened 2022-06-15
Problem: cancel gravity transfer integration test fail accationally


yihuang:
https://github.com/crypto-org-chain/cronos/runs/6878397070?check_suite_focus=true

because the cancel call could fail occasionally because the transfer is already included in a batch.
@thomas-nguy any idea how to assert this more deterministically?

thomas-nguy:
batch is created every 10 blocks.

- one solution is to increase block production time and make sure that send_to_eth tx in inserted in block 10X + 1 but this is very tricky to do
- another solution is to do send_to_eth and cancel_batch in batch transaction but we won't be able to verify the amount deducted and refunded

thomas-nguy:
I guess if it keeps failing, we will opt for solution 2?

yihuang:
but do it in batch, you can't get the packet id?

thomas-nguy:
yeah the txId is generated in order, if tests run sequentially within test_gravity or clean the state before each test, we will be able to get the id deterministically but is it the case in python?

yihuang:
> yeah the txId is generated in order, if tests run sequentially within test_gravity or clean the state before each test, we will be able to get the id deterministically but is it the case in python?

I see, if you fetch the id at the beginning and increase by one, it should work.

PR #564, opened 2022-06-29
Problem: cancel gravity transfer integration test fail accationally

mmsqe:
Closes: https://github.com/crypto-org-chain/cronos/issues/553

# PR Checklist:

### DIFF
--- a/integration_tests/test_gravity.py
+++ b/integration_tests/test_gravity.py
@@ -258,9 +258,12 @@ def test_gravity_transfer(gravity):
 
     if gravity.cronos.enable_auto_deployment:
         crc21_contract = None
+
         def local_check_auto_deployment():
             nonlocal crc21_contract
-            crc21_contract = check_auto_deployment(cli, denom, cronos_w3, recipient, amount)
+            crc21_contract = check_auto_deployment(
+                cli, denom, cronos_w3, recipient, amount
+            )
             return crc21_contract
 
         wait_for_fn("send-to-crc21", local_check_auto_deployment)
@@ -446,7 +449,7 @@ def test_gravity_cancel_transfer(gravity):
         balance = erc20.caller.balanceOf(ADDRS["validator"])
         assert balance == 100000000000000000000000000
         amount = 1000
-        
+
         current_block = int(cli.status()["SyncInfo"]["latest_block_height"])
         batch_block = 10
         diff_block = batch_block - current_block % batch_block
@@ -456,12 +459,15 @@ def test_gravity_cancel_transfer(gravity):
         community = HexBytes(ADDRS["community"])
         send_to_cosmos(gravity.contract, erc20, community, amount, KEYS["validator"])
         assert erc20.caller.balanceOf(ADDRS["validator"]) == balance - amount
-        
+
         denom = f"gravity{erc20.address}"
         crc21_contract = None
+
         def local_check_auto_deployment():
             nonlocal crc21_contract
-            crc21_contract = check_auto_deployment(cli, denom, cronos_w3, community, amount)
+            crc21_contract = check_auto_deployment(
+                cli, denom, cronos_w3, community, amount
+            )
             return crc21_contract
 
         wait_for_fn("send-to-crc21", local_check_auto_deployment)
--- a/integration_tests/test_gravity.py
+++ b/integration_tests/test_gravity.py
@@ -428,6 +428,7 @@ def test_direct_token_mapping(gravity):
 def test_gravity_cancel_transfer(gravity):
     if gravity.cronos.enable_auto_deployment:
         geth = gravity.geth
+        cli = gravity.cronos.cosmos_cli()
         cronos_w3 = gravity.cronos.w3
 
         # deploy test erc20 contract
@@ -445,13 +446,17 @@ def test_gravity_cancel_transfer(gravity):
         balance = erc20.caller.balanceOf(ADDRS["validator"])
         assert balance == 100000000000000000000000000
         amount = 1000
+        
+        current_block = int(cli.status()["SyncInfo"]["latest_block_height"])
+        batch_block = 10
+        diff_block = batch_block - current_block % batch_block
+        wait_for_new_blocks(cli, diff_block)
 
         print("send to cronos crc21")
         community = HexBytes(ADDRS["community"])
         send_to_cosmos(gravity.contract, erc20, community, amount, KEYS["validator"])
         assert erc20.caller.balanceOf(ADDRS["validator"]) == balance - amount
         
-        cli = gravity.cronos.cosmos_cli()
         denom = f"gravity{erc20.address}"
         crc21_contract = None
         def local_check_auto_deployment():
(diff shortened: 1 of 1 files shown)


thomas-nguy:
As discussed https://github.com/crypto-org-chain/cronos/issues/553 

maybe a better way would be to send the "send_to_chain" and "cancel_send_to_chain" in the same transaction so that it wont have time to be included in a batch.

The problematic part is that "cancel_send_to_chain" rely on the tx_id returned in `send_to_chain" receipt.

This id is incremented for every "send_to_chain" or "SendToEthereum" call. Unfortunately there is no way to query it directly from the chain so one recommendation would be to keep a global variable that would increment everytime we call "send_to_chain" or "SendToEthereum" in a test?

Anyways this is the recommended way to be sure that we won't have a concurrent issue. If the current way works. then I guess it is also fine.

mmsqe:
> As discussed #553

thanks the detail, I can try build_batch_tx(cronos_w3, cli, [tx, canceltx]), but need sure sequence, and seems need replace balance check with emitted event amount check

thomas-nguy:
yeah if you build  "send_to_chain" and "cancel_send_to_chain" into a single batch, maybe we can modify the check that the balance is the same before and after batch

but it is less powerful than previous check

yihuang:
> yeah if you build "send_to_chain" and "cancel_send_to_chain" into a single batch, maybe we can modify the check that the balance is the same before and after batch
> 
> but it is less powerful than previous check

it seems there's no query handler to fetch the next batch id, so impossible to implement the solution 2?

mmsqe:
> > yeah if you build "send_to_chain" and "cancel_send_to_chain" into a single batch, maybe we can modify the check that the balance is the same before and after batch
> > but it is less powerful than previous check
> 
> it seems there's no query handler to fetch the next batch id, so impossible to implement the solution 2?

I didnt find query for [id](https://github.com/crypto-org-chain/gravity-bridge/blob/v2.0.0-cronos/module/x/gravity/types/key.go#L47) but thinking could append `__CronosSendToChainResponse` in [abi](https://github.com/crypto-org-chain/cronos/blob/main/integration_tests/test_gravity.py#L38) to make event filter work

yihuang:
> ```
> This id is incremented for every "send_to_chain" or "SendToEthereum" call. Unfortunately there is no way to query it directly on chain so one way to do it would be to keep a global variable that would increment everytime we call "send_to_chain" or "SendToEthereum" in a test.
> ```
> 
> There is no query, that is why I also suggested to maintain a global counter that will increase everytime we call those functions in ours tests. As long as we are they are not running parallelly, that could work

that may be a little bit uglier, considering the current solution already works, almost.

mmsqe:
> that may be a little bit uglier, considering the current solution already works, almost.

yea, seems 1st opt easier compare with [2nd](https://github.com/mmsqe/cronos/commit/f558d4c387bec07d8ce9dc9c24cfa2b69c90e63d) test batch with global counter

thomas-nguy:
okay it if works, I guess it is also fine.

integration_tests/test_gravity.py
```diff
@@ -440,52 +451,32 @@ def test_gravity_cancel_transfer(gravity):
         assert balance == 100000000000000000000000000
         amount = 1000
 
+        current_block = int(cli.status()["SyncInfo"]["latest_block_height"])
```
thomas-nguy, line 454:
```suggestion
        # batch are created every 10 blocks, wait til block number reaches a multiple of 10 to lower the chance to have the transaction include in the batch right away
        current_block = int(cli.status()["SyncInfo"]["latest_block_height"])
```

integration_tests/test_gravity.py
```diff
@@ -440,52 +451,32 @@ def test_gravity_cancel_transfer(gravity):
...
-        def check_fund():
-            v = crc21_contract.caller.balanceOf(ADDRS["community"])
-            return v == amount
+            return crc21_contract
 
-        wait_for_fn("send-to-ethereum", check_fund)
+        wait_for_fn("send-to-crc21", local_check_auto_deployment)
 
         # send it back to erc20
         tx = crc21_contract.functions.send_to_chain(
```
thomas-nguy, line 477: Better move the wait function here.  The transaction that we want to avoid being included in a batch is "send_to_chain"

thomas-nguy (approved):
lgtm

integration_tests/test_gravity.py
```diff
@@ -441,51 +452,35 @@ def test_gravity_cancel_transfer(gravity):
...
-                ):
-                    return log.data
-            return "0x0000000000000000000000000000000000000000000000000000000000000000"
-
-        wait_for_fn("send-to-crc20", check_auto_deployment)
-
-        def check_fund():
-            v = crc21_contract.caller.balanceOf(ADDRS["community"])
-            return v == amount
+            return crc21_contract
```
yihuang, line 467:
```suggestion
            return crc21_contract.caller.balanceOf(recipient) == amount
```

It should return a boolean here?

integration_tests/test_gravity.py
```diff
@@ -441,51 +452,35 @@ def test_gravity_cancel_transfer(gravity):
...
+
+        wait_for_fn("send-to-crc21", local_check_auto_deployment)
 
-        wait_for_fn("send-to-ethereum", check_fund)
```
yihuang, line 483: this wait call is removed, but it seems not replaced with a new one?

integration_tests/test_gravity.py
mmsqe: need return contract back for [test](https://github.com/mmsqe/cronos/blob/bb9ffe112678f9e0b6c33ebdf289108b78501ba2/integration_tests/test_gravity.py#L273) to use

integration_tests/test_gravity.py
mmsqe: [check_auto_deployment](https://github.com/crypto-org-chain/cronos/blob/db79ab90c7a2108c26052f35440095a2e7cffbb4/integration_tests/test_gravity.py#L464) already check balance for [recipient](https://github.com/crypto-org-chain/cronos/blob/db79ab90c7a2108c26052f35440095a2e7cffbb4/integration_tests/test_gravity.py#L444) which is [community](https://github.com/crypto-org-chain/cronos/blob/db79ab90c7a2108c26052f35440095a2e7cffbb4/integration_tests/test_gravity.py#L480)

integration_tests/test_gravity.py
yihuang: yeah, we should do following procedure for maximum stability:
- send-to-cosmos, and wait for balance
- wait for the edge block
- send-to-chain
- cancel

yihuang (changes requested):
https://github.com/crypto-org-chain/cronos/pull/564/files/8eb3ec0fc9db2236cd73a97ae3db78e99231f86a#r910668135

integration_tests/test_gravity.py
yihuang: it's already the case

yihuang (approved):
LGTM

PR #600, opened 2022-07-25
Problem: bi-directional token conversion is not implemented

thomas-nguy:

Solution: Implement ADR-008

Todo: add test
# PR Checklist:


contracts/src/ModuleCRC21.sol
```diff
@@ -1,34 +1,104 @@
-pragma solidity ^0.6.8;
+pragma solidity ^0.6.1;
```
thomas-nguy, line 1: needs to update compiler to latest

contracts/src/ModuleCRC21.sol
```diff
@@ -1,34 +1,104 @@
-pragma solidity ^0.6.8;
```
thomas-nguy, line 1: @FrancoCRO could you help review this contract 🙏

yihuang (changes requested):
integration test need to be updated, the code changes looks good.

x/cronos/keeper/evm.go
```diff
@@ -95,47 +94,84 @@ func (k Keeper) ConvertCoinFromNativeToCRC20(ctx sdk.Context, sender common.Addr
...
+		err = k.bankKeeper.SendCoinsFromAccountToModule(ctx, sdk.AccAddress(sender.Bytes()), types.ModuleName, sdk.NewCoins(coin))
+		if err != nil {
+			return err
+		}
+		err = k.bankKeeper.BurnCoins(ctx, types.ModuleName, sdk.NewCoins(coin))
+		if err != nil {
+			return err
+		}
+		// unlock crc tokens
+		_, err = k.CallModuleCRC21(ctx, contract, "transfer_by_cronos_module", sender, coin.Amount.BigInt())
```
adu-crypto, line 110:
```solidity
    function transfer_by_cronos_module(address addr, uint amount) public {
        require(msg.sender == module_address);
        unsafe_transfer(addr, module_address, amount);
    }
```
it seems `transfer_by_cronos_module` is used to lock crc21 tokens?

yihuang: should be `transfer_to_cronos_module`?

adu-crypto: I mean the comment says `unlock crc tokens`, so I think we should unlock tokens by calling `transfer_from_cronos_module`.

thomas-nguy: yes it should be `transfer_from_cronos_module`

yihuang (changes requested):
Code changes looks good, can you add an integration test case to boost confidence?

thomas-nguy:
integration tests for ibc added there
https://github.com/crypto-org-chain/cronos/pull/609/files

yihuang (approved):
LGTM

contracts/src/ModuleCRC21.sol
```diff
@@ -1,34 +1,104 @@
...
+        emit Burn(addr, amount);
+    }
+
+    // unsafe_transfer transfer tokens without user's approval and authentication, used internally
+    function unsafe_transfer(address src, address dst, uint amount) internal {
+        require(balanceOf[src] >= amount, "ds-token-insufficient-balance");
+        balanceOf[src] = sub(balanceOf[src], amount);
+        balanceOf[dst] = add(balanceOf[dst], amount);
+        emit Transfer(src, dst, amount);
+    }
```
FinnZhangCrypto, line 103: Based on the whole code, looks all good to me. Not sure if the DSToken code is the same as the current ModuleCRC20 token in `cronos-token-init-launch` repo, if so then all good.

Additional question about the `isSource` flag:
- in the current version, this flag can only be initialized once in constructor when it was deployed. It means we cannot switch contract model to the previous version of `ModuleCRC20`. Is that an expected feature? Can you provide more details?
- I'm not sure if the module_address will perform a burn operation in actual script(guess it should do to keep consistence with previous version?). And please ensure the actual burn operation() after triggered the corresponding `__CronosSendToIbc` or `__CronosSendToChain` event.
- Can you explain what this module_address actually is?

FinnZhangCrypto: Or I guess this module_address will transfer or burn the corresponding tokens according to the actual feature(to chain or to ibc). Like this below:
- to ibc: transfer tokens from user to module_address, then module_address burn tokens to keep consistence with previous version.
- to chain: transfer tokens from user to module_address, then module_address transfer to Gravity Bridge to the target chain(e.g. Ethereum).

Correct me if I'm wrong. Thanks!

FinnZhangCrypto: Or I guess you will use module_address to do different operations according to the actual use?
Based on the previous `ModuleCRC20` contract, all tokens will be burn when sending tokens:
- to ibc: 2 kinds of methods to do that, transfer to module_address then burn / transfer to module_address and directly to ibc(when getting back, directly transfer back to user, doesn't need to mint)
- to chain: 2 kinds of methods to do that, transfer to module_address then burn(mint tokens on GB side, when getting back, directly burn GB side and mint in this CRC21 side) / the same as above.

thomas-nguy: to clarify about the source boolean.

Source define is the contract token represent an asset originated from the chain (cronos) like VVS, FERRO etc or comes from another chain (ETH, WBTC, ATOM) 

It should be defined during initialisation and should probably not change.

 
`module_address` is a special address use by the module when performing an evm operation. The module will invoke the smart contract method bypassing the orignal transaction flow and the "msg.sender" address will be set to `module_address`. It holds certains priviledge in the contract such as be able to mint or burn tokens freely or move tokens 

The reason we are not "burning/minting" tokens for `source` contract but move them instead is that we want to preserve the total supply of the tokens (calculated by total_supply()) 
For non-source tokens, burning is okay because the total supply is defined on the original contract (not in cronos)

yihuang: > Can you explain what this module_address actually is?

`module_address` is only used in:
```
require(msg.sender == module_address);
```
which guarantees it can only be called by chain native code, rather than anyone else.

PR #601, opened 2022-07-25
Problem: Copyright is missing attribution

calvinaco:
Solution: Update LICENSE file


yihuang (approved):

LICENSE
```diff
@@ -1,4 +1,5 @@
-  Copyright 2021-present
+
+   Copyright 2021-present Crypto.org
```
thomas-nguy, line 2: btw why don't we use cronos labs?

tomtau: cronos labs are used on the ecosystem projects of cronos labs: https://github.com/cronos-labs

PR #615, opened 2022-08-03
Problem: Missing changes to ADR007 for generic event arguments

calvinaco:
Solution: Revise the ADR with the new proposal

# PR Checklist:


docs/architecture/adr-007.md
```diff
@@ -10,88 +13,101 @@ The current way to define those events is not flexible: events are hardcoded, pa
...
+### __CronosSendToChainV2 Event
 
-```
-# Trigger an action after contract execution which only involve the sender account
-__CosmosCall(string functionName, address sender,  byte[]args)
+#### Changes
 
-# Trigger an action after contract execution that require moving asset from the smart contract "account"
-__CosmosCallAfterPayment(string functionName, address payer, uint amount, byte[]args)
+- Introduce event `__CronosSendToChainV2`, successor of event `__CronosSendToChain`;
```
thomas-nguy, line 24: reposting my comment,

Why adding a V2? we can keep the same name

docs/architecture/adr-007.md
```diff
@@ -10,88 +13,101 @@ The current way to define those events is not flexible: events are hardcoded, pa
...
+- Add new argument `string channelId` to accept destination channel Id;
```
thomas-nguy, line 25: what the use of channelID?

I think its better to separate CronosSendToIBC and CronosSendToChain, otherwise we would have the same security concern as the previous proposed ADR

docs/architecture/adr-007.md
```diff
@@ -10,88 +13,101 @@ The current way to define those events is not flexible: events are hardcoded, pa
...
-     emit __CosmosCallAfterPayment("CronosSendToIbc", msg.sender, amount, args);
- }
+#### Signature
 
+```solidity
+// Current
+event __CronosSendToChain(address sender, address recipient, uint256 amount, uint256 bridge_fee, uint256 chain_id);
+
+// Proposal
+event __CronosSendToChain(address sender, address recipient, uint256 amount, uint256 bridge_fee, string chain_id, extraData bytes);
```
thomas-nguy, line 60: bridge_fee can be also encapsulated in extraData since its not clear what would be the format

docs/architecture/adr-007.md
```diff
@@ -10,88 +13,101 @@ The current way to define those events is not flexible: events are hardcoded, pa
...
 ```
 
-In this example byte[] args should contains the encoded recipient address from `event __CronosSendToIbc(address sender, string recipient, uint256 amount)` but in the future, if we needs to add extra parameters such as `fee_amount` or `fee_recipient` we will be able to use the same method in the smart contract, but encode those extra parameters in args. 
+#### Description
 
-On the chain side, the events parameters are decoded and uses to call a specific actions. 
+- Chain ID argument is chagned from unsigned integer to string. This allow sending token to destination chains other than Ethereum (Gravity Bridge).
```
thomas-nguy, line 65: for non ethereum chain, should be redefine new types of event instead?

docs/architecture/adr-007.md
calvinaco: This is a design decision only.

Use the same event:
- Two format of the same event exist;
- Chain logic has to perform the argument existence and type check at runtime, doable but maybe more error prone

Upgrade event
- Event format is fixed, i.e. argument type remains unchanged;
- Separated chain logic and maybe less error prone

So I decide to favour the readability and clear chain logic.

docs/architecture/adr-007.md
calvinaco: Actually it's a typo. This is the `__CronosSendToIBC` event

docs/architecture/adr-007.md
calvinaco: Sure, let me update it.

docs/architecture/adr-007.md
calvinaco: For some reason my last comment was gone, let me re-type it

When I first look at `__CronosSendToChain` I thought it was intended for sending to any arbitrary chains. I could think of a few directions for this event:

1. A generic bridge event, pretty much the same for now that includes sender, recipient, amount (with denom), destination + extra data for chain/protocol specific arguments
2. By bridge protocol, `__CronosSendToGravityBridge` just like what `__CronosSendToIBC` is doing. Caveat is token still need to migrate when we add new protocols
3. By destination chain, `__CronosSendToEthereum`, similar to 2.

thomas-nguy: maybe we could rename to `__CronosSendToEVMChain`

For non evm chain, it is not clear what would be the flow or the parameters at the moment, not sure if we can design it now, unless we encapsulate all the parameters in []args except the one used by the contract

calvinaco: > maybe we could rename to __CronosSendToEVMChain
And for EVM chain the underlying protocol is via Gravity Bridge right?

thomas-nguy: yes gravity bridge is compatible with any EVM compatible chain.

docs/architecture/adr-007.md
thomas-nguy: since `SendToIBC` to reintroduce in CRC21, I think we could keep the same name but having different parameters

CRC20 will be marked as deprecated and ask people to uses CRC21

We will still support both events, in the hook level, we could simply check the arguments

thomas-nguy:
other that that, lgtm

docs/architecture/adr-007.md
calvinaco: Ok I will update this section.

docs/architecture/adr-007.md
calvinaco: Updated

docs/architecture/adr-007.md
```diff
@@ -10,88 +13,104 @@ The current way to define those events is not flexible: events are hardcoded, pa
...
-- **functionName** corresponds to the equivalent hardcoded eventName such as `CronosSendToIBC` for the event `__CronosSendToIbc` or `CronosSendToChain` for the event `__CronosSendToChain`
-- **payer** is the address which make the payment, usually equal to `msg.sender`
-- **amount** corresponds to the `amount` that the payer has paid (either by burning or sending asset to the SC)
-- **args** corresponds to the argument in order of the event removing the `sender` and `amount` from the parameters encoded using the abi event pack
+```solidity
+// Current
+event __CronosSendToIbc(address sender, string recipient, uint256 amount);
 
+// Proposal
+event __CronosSendToIbcV2(address sender, string recipient, uint256 amount, string channelId, bytes extraData);
```
thomas-nguy, line 34: remove v2?

docs/architecture/adr-007.md
```diff
@@ -10,88 +13,104 @@ The current way to define those events is not flexible: events are hardcoded, pa
...
-```
+#### Signature
 
-- **functionName** corresponds to the equivalent hardcoded eventName such as `CronosSendToIBC` for the event `__CronosSendToIbc` or `CronosSendToChain` for the event `__CronosSendToChain`
-- **payer** is the address which make the payment, usually equal to `msg.sender`
-- **amount** corresponds to the `amount` that the payer has paid (either by burning or sending asset to the SC)
-- **args** corresponds to the argument in order of the event removing the `sender` and `amount` from the parameters encoded using the abi event pack
+```solidity
+// Current
+event __CronosSendToIbc(address sender, string recipient, uint256 amount);
```
chrisCryptoCom, line 31: Shouldn't the recipient be of type `address` ?

docs/architecture/adr-007.md
```diff
@@ -10,88 +13,104 @@ The current way to define those events is not flexible: events are hardcoded, pa
...
-     emit __CosmosCallAfterPayment("CronosSendToIbc", msg.sender, amount, args);
- }
+#### Signature
 
+```solidity
+// Current
+event __CronosSendToChain(address sender, address recipient, uint256 amount, uint256 bridge_fee, uint256 chain_id);
+
+// Proposal
+event __CronosSendToEVMChain(address sender, address recipient, uint256 amount, string chain_id, bytes extraData);
```
chrisCryptoCom, line 61: Maybe you'd want to specify some of the event arguments as `indexed` so they are treated as searchable log hashses instead of simple data ?
https://ethereum.stackexchange.com/questions/8658/what-does-the-indexed-keyword-do

docs/architecture/adr-007.md
JohnYuCrypto: I think it's `string` because it can be a chain other than evm-compatible one?

docs/architecture/adr-007.md
```diff
@@ -10,88 +13,104 @@ The current way to define those events is not flexible: events are hardcoded, pa
...
 ```
 
-In this example byte[] args should contains the encoded recipient address from `event __CronosSendToIbc(address sender, string recipient, uint256 amount)` but in the future, if we needs to add extra parameters such as `fee_amount` or `fee_recipient` we will be able to use the same method in the smart contract, but encode those extra parameters in args. 
+#### Description
 
-On the chain side, the events parameters are decoded and uses to call a specific actions. 
+- Event is renamed to make it self-descritive on the purpose.
+- Chain ID argument is chagned from unsigned integer to string. This allow sending token to destination chains other than Ethereum (Gravity Bridge).
```
JohnYuCrypto, line 67: changed not chagned

docs/architecture/adr-007.md
JohnYuCrypto: Agreed, max indexed arguments number in event is 3, maybe set first 3 to `indexed`

docs/architecture/adr-007.md
calvinaco, line 67:
```suggestion
- Chain ID argument is changed from unsigned integer to string. This allow sending token to destination chains other than Ethereum (Gravity Bridge).
```

docs/architecture/adr-007.md
calvinaco, line 34:
```suggestion
event __CronosSendToIbc(address sender, string recipient, uint256 amount, string channelId, bytes extraData);
```

docs/architecture/adr-007.md
calvinaco: Yes, that's correct. The address will be an arbitrary Cosmos address

docs/architecture/adr-007.md
calvinaco: Thanks! Will add the index

calvinaco: On the other hand, @chrisCryptoCom @JohnYuCrypto , I am not an expert on solidity, may I know if `bytes` is a suitable type here? I am looking for a type that can be dynamic in size to accommodate future arguments. But at the same time I want to understand the trade off, if there's any. Some I could think of
- Will it be much harder for smart contract to process a `bytes` as it is dynamic?
- Is it really dynamic as in it can accept inputs at different lengths? Is there a limitation?
- Library support, does library like web3 capable to prepare this field?
Thanks

chrisCryptoCom: _"Will it be much harder for smart contract to process a bytes as it is dynamic?" - To process the data the contract will have to know the types / structure of the data that was encoded in the first place. See solidity docs [here](https://docs.soliditylang.org/en/v0.8.15/units-and-global-variables.html?highlight=abi.decode#abi-encoding-and-decoding-functions)

- "Is it really dynamic as in it can accept inputs at different lengths? Is there a limitation?" - Yes it is dynamically sized and you can encode arbitrary amounts and types of data. (see above link for methods to encode data from solidity)

- "Library support, does library like web3 capable to prepare this field?"  [web3](https://web3js.readthedocs.io/en/v1.7.5/web3-eth-abi.html) and [ethers](https://docs.ethers.io/v5/api/utils/abi/coder/) both have helper methods for decoding / encoding data to / from bytes.

JohnYuCrypto: Limitation of the size of the bytes passed in as function argument and emitted in event should only be restricted by gas limit.

adu-crypto: Is this data used for arbitrary command execution? From my personal experience, arbitrary execution could be vulnerable considering security(referring to [polynetwork hack](https://slowmist.medium.com/the-analysis-and-q-a-of-poly-network-being-hacked-8112a35beb39) and many other defi exploitation). Could we use a manageable whitelist if it's used fro arbitrary command execution?

thomas-nguy (approved):
LGTM 

However one thing which is not specified is how to serialized the args to byte[] assuming args will be a struct or object.

Since most of our users would be using javascript or python, we have quite some options there (JSON, protobuf etc)

Any suggestion @yihuang @tomtau

docs/architecture/adr-007.md
```diff
@@ -10,88 +13,104 @@ The current way to define those events is not flexible: events are hardcoded, pa
...
+- Possibility to remove or add parameters to existing actions (if we upgrade ibc-go or gravity module for example);
+- Reduce number of upgrade or migrations requirement to existing token contracts.
 
 ### Negative
-- Might have some security concerns if those actions are not correctly "bounded" by a defined scope (smart contract can only move its own asset, action can only be performed from sender account)
-- Not possible to restrict bundled features for smart contract developers once they implement the generic events
 
+- Might have some security concerns if those arguments are not correctly "bounded" by a defined scope (smart contract can only move its own asset, action can only be performed from sender account);
+- Extra data defined in byte array will make the contract less readable.
+- Manpulation of argument inside a dynamic-sized byte array (bytes) is not trivial;
```
tomtau, line 108:
```suggestion
- Manipulation of argument inside a dynamic-sized byte array (bytes) is not trivial;
```

docs/architecture/adr-007.md
```diff
@@ -10,88 +13,104 @@ The current way to define those events is not flexible: events are hardcoded, pa
...
+- Extra complexity for smart contract to restrict future bridge features because the future arguemnt formats will not be known at deploy time.
```
tomtau, line 109:
```suggestion
- Extra complexity for smart contract to restrict future bridge features because the future argument formats will not be known at deploy time.
```

tomtau:
> However one thing which is not specified is how to serialized the args to byte[] assuming args will be a struct or object which contains multiple parameters.

if I understand it correctly, I think that's purposefully left open, i.e. left for dapps to converge on something that makes sense in their future context.

thomas-nguy:
```
if I understand it correctly, I think that's purposefully left open, i.e. left for dapps to converge on something that makes sense in their future context.
```

Technically it cannot be left open for dapps to decide on their standard because the data needs to be deserialize on-chain, depending on our implementation, dapps would need to follow the same serialization rule

docs/architecture/adr-007.md
calvinaco: Hello @adu-crypto , it is not arbitrary command execution but a bounded execution. However, some of the elements in the hack are still valuable to us:

1. The source chain did not check the initiated cross-chain operation.
2. The target chain did not check the parsed target call contract and call parameters.

Our EVM hooks should decode and validate the arguments based on version, but if the validation has flaws, then it may still open to some attack vectors. So this has to be carefully implemented.

calvinaco: @chrisCryptoCom @JohnYuCrypto @adu-crypto thanks again for the comments. It has helped the ADR a lot.

After today's technical meeting we have one more question: is there any recommendation on the format and serialization for the extra arguments to bytes? I think the implementer of this ADR could definitely come up with something but we would like to see if there's any convention out there.

Thanks!

adu-crypto: I believe the `ExtraData` is used by the native code outside of EVM to trigger specific hook actions(correct me if I'm wrong), so it's native code's job to decode the `ExtraData`. So I think native code's deserialization should be compatible with EVM's serialization. 
According to @chrisCryptoCom 's comment and [ethers' examples](https://docs.ethers.io/v5/api/utils/abi/coder/#AbiCoder-x), both raw data and concrete types should be provided for the `ExtraData` so that the native code could decode the raw data correctly. Or maybe we should agree on both EVM side and native code side about the concrete types of the `ExtraData`.

docs/architecture/adr-007.md
```diff
@@ -10,88 +13,104 @@ The current way to define those events is not flexible: events are hardcoded, pa
...
+- The new event formats add support to arbitrary destination chains according to the capability of the chain.
+- It is up to the contract to define restriction on the destination, if needed.
 
- 
-With this generic event format, it is possible to add and remove the parameters for those action in the future or even add new actions while "upgrading" the chain. Smart contract won't be force to migrate to a new contract to benefit those functionnalities.
+### Extra Arguments Guidelines
+
+This is a guideline to future bridge features on leveraging the `bytes extraData` arguement:
+
+1. The first 2 bytes are reserved be a version number, to help the chain logic to recognize the format of elements in `extraData` array.
```
adu-crypto, line 80:
```suggestion
1. The first 2 bytes are reserved as a version number, to help the chain logic to recognize the format of elements in `extraData` array.
```

thomas-nguy:
some checks are failing

calvinaco:
> some checks are failing

It's from another file. Fixing it here: https://github.com/crypto-org-chain/cronos/pull/679 and we can merge this PR I believe.

adu-crypto:
@calvinaco  should we clarify that the `extraData` should be encoded outside of the evm call and passed into the contract call as the input, otherwise if the contract encoded the `extraData` inside contract and we upgraded the chain logic about `extraData` encoding/decoding, contract devs still need to upgrade the contract.

calvinaco:
> @calvinaco should we clarify that the `extraData` should be encoded outside of the evm call and passed into the contract call as the input, otherwise if the contract encoded the `extraData` inside contract and we upgraded the chain logic about `extraData` encoding/decoding, contract devs still need to upgrade the contract.

Yes, I think this can be added as a guideline but not enforced. The contract developer may still want to do some tricks by constructing the argument by themselves, as long as their contract is upgradable this can be managed.

calvinaco:
@adu-crypto @yihuang If there's no other concerns I think we can merge this.

PR #770, opened 2022-11-15
Problem: existing crc20 are not compatible with crc21 standard

thomas-nguy:
**Open this PR just for discussion**

Pro of this methods :

- No need to use a wrapper, which remove the burden to deal with wrap tokens

Cons:

- Need an extra approval for transferring to Ethereum
- The totalSupply should be queried to the proxy contract, not the actual CRC20 contract. Not sure if it can be an issue for existing dapps?

### DIFF
--- a/contracts/src/ModuleCRC20Proxy.sol
+++ b/contracts/src/ModuleCRC20Proxy.sol
@@ -3,39 +3,57 @@ pragma solidity ^0.6.8;
 import "ds-math/math.sol";
 import "./ModuleCRC20.sol";
 
+contract ProxyAuthority {
+    address proxyAddress;
+
+    constructor(address _proxyAddress) public {
+        proxyAddress = _proxyAddress;
+    }
+
+    function canCall(
+        address src, address dst, bytes4 sig
+    ) public view returns (bool) {
+        if (msg.sender == proxyAddress) {
+            return true;
+        }
+
+        return false;
+    }
+}
+
 contract ModuleCRC20Proxy is DSMath {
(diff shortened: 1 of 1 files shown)


contracts/src/ModuleCRC20Proxy.sol
```diff
@@ -0,0 +1,58 @@
...
+
+    /**
+     * @dev Returns the number of tokens currently in circulation
+     * This value should reflect the number of token which are locked in the gravity bridge
+     * corresponding to the crc20 token.
+     * Note that it is not an accurate number but only an estimation as they may be latency between the ethereum network
+     * and cronos network.
+	 *
+	 */
+    function totalSupply() public view virtual override returns (uint256) {
```
thomas-nguy, line 55: totalSupply should be queried from the proxyContract, not the actual CRC20 contract. Could it be an issue for existing Dapps? @yijia

contracts/src/ModuleCRC20Proxy.sol
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    /**
+        Evm hooks functions
+    **/
+
+    // send to another chain through gravity bridge
+    function send_to_evm_chain(address recipient, uint amount, uint chain_id, uint bridge_fee, bytes calldata extraData) external {
+        // transfer back the token to the proxy account
+        crc20Contract.transferFrom(msg.sender, address(this), add(amount,bridge_fee));
```
thomas-nguy, line 38: requires an extra approval

Issue #704, opened 2022-09-19
Problem: there's no compact historical state storage


yihuang:
A solution for: https://github.com/crypto-org-chain/cronos/issues/451
Since we already have the state streaming in SDK(ADR 038), it's possible to implement the idea in Cronos itself.
Just implement a streaming service that store the state changes in a kv db, the format is explained [here](https://github.com/cosmos/cosmos-sdk/issues/13317).

yihuang:
### File streamer integration
- [x] 0.9.x https://github.com/crypto-org-chain/cronos/pull/721

yihuang:
closed by #791

PR #791, opened 2022-12-14
Problem: no compact storage for archive node

yihuang:
Closes: #704 

Solution:
- patch iavl to support extracting state changes
- implement versiondb using rocksdb user-defined timestamp feature
- ~~now librocksdb is always required to build cronosd~~
  versiondb is only enabled if binary is built with rocksdb backend.
- implement cli commands to manage the migration process:

```
$ ./build/cronosd changeset
dump and manage change sets files and ingest into versiondb

Usage:
  cronosd changeset [command]

Available Commands:
  convert-to-sst Convert change set files to versiondb/rocksdb sst files, which can be ingested into versiondb later
  dump           Extract changesets from iavl versions, and save to plain file format
  ingest-sst     Ingest sst files into versiondb
  print          Pretty-print the content of change set file
  to-versiondb   Feed change set files into versiondb
  verify         Replay the input change set files in order to rebuild iavl tree in memory and output root hash, user can compare the root hash against the on chain hash
```

- [Review Guides](#issuecomment-1403201363)

- [VersionDB Documentation](https://github.com/crypto-org-chain/cronos/blob/45162664a0f7bf25f4a3c14db01baf45d737a98a/versiondb/README.md)

- [The Future Plan for IAVL Tree Itself](https://github.com/crypto-org-chain/cronos/blob/45162664a0f7bf25f4a3c14db01baf45d737a98a/versiondb/memiavl/README.md)

- [Performance Numbers in Testnet Run](#issuecomment-1403505389)

# PR Checklist:


cmd/cronosd/cmd/changeset.go
```diff
@@ -0,0 +1,687 @@
...
+					return err
+				}
+				latestVersion, err := tree.LazyLoadVersion(0)
+				if err != nil {
+					return err
+				}
+				endVersion = latestVersion + 1
+			}
+
+			tasksChan := make(chan *dumpTask, 2^32)
```
tomtau, line 128: it looks like, this meant to be 2 << 32? 2^32==34

tomtau (commented):
will there be any fallback for other db backends?

cmd/cronosd/cmd/changeset.go
yihuang: ah, yes

yihuang:
> will there be any fallback for other db backends?

that would be the original proposal, if there are any interests for that.
user can still specify different db backends for tm and iavl db though, just the versiondb will be always rocksdb.

cmd/cronosd/cmd/changeset.go
yihuang: it's reallocate the buffer, so I change it to 2048, should be big enough.

go.mod
```diff
@@ -200,10 +204,15 @@ replace (
...
 	github.com/gogo/protobuf => github.com/regen-network/protobuf v1.3.3-alpha.regen.1
 
+	github.com/linxGnu/grocksdb => github.com/yihuang/grocksdb v1.7.11-0.20221223014353-a4fa14d1759b
+
 	// TODO: remove when gravity update dependencies
 	github.com/peggyjv/gravity-bridge/module/v2 => github.com/crypto-org-chain/gravity-bridge/module/v2 v2.0.1-0.20220815102151-48275db7e1ee
 	github.com/tendermint/tendermint => github.com/tendermint/tendermint v0.34.24-0.20221110131553-ec471ba27efd
 
+	// https://github.com/yihuang/tm-db/tree/rocksdb-v7
+	github.com/tendermint/tm-db => github.com/yihuang/tm-db v0.0.0-20221223014825-64db9113ab50
```
yihuang, line 214: https://github.com/tendermint/tm-db/pull/312

go.mod
```diff
@@ -200,10 +204,15 @@ replace (
 	github.com/gin-gonic/gin => github.com/gin-gonic/gin v1.7.0
 	github.com/gogo/protobuf => github.com/regen-network/protobuf v1.3.3-alpha.regen.1
 
+	github.com/linxGnu/grocksdb => github.com/yihuang/grocksdb v1.7.11-0.20221223014353-a4fa14d1759b
```
yihuang, line 207: https://github.com/linxGnu/grocksdb/pull/100
fix easier build.

go.mod
```diff
@@ -200,10 +204,15 @@ replace (
...
+
 	// TODO: remove when gravity update dependencies
 	github.com/peggyjv/gravity-bridge/module/v2 => github.com/crypto-org-chain/gravity-bridge/module/v2 v2.0.1-0.20220815102151-48275db7e1ee
 	github.com/tendermint/tendermint => github.com/tendermint/tendermint v0.34.24-0.20221110131553-ec471ba27efd
 
+	// https://github.com/yihuang/tm-db/tree/rocksdb-v7
+	github.com/tendermint/tm-db => github.com/yihuang/tm-db v0.0.0-20230103092224-3f21822ed03e
```
yihuang, line 214: https://github.com/tendermint/tm-db/pull/312
migrate to grocksdb to support rocksdb v7.

go.mod
```diff
@@ -192,6 +194,9 @@ replace (
 	// Ref: https://forum.cosmos.network/t/ibc-security-advisory-dragonberry/7702
 	github.com/confio/ics23/go => github.com/confio/ics23/go v0.9.0
 	github.com/cosmos/cosmos-sdk => github.com/cosmos/cosmos-sdk v0.46.7
+
+	// https://github.com/yihuang/iavl/tree/dump_changeset
+	github.com/cosmos/iavl => github.com/yihuang/iavl v0.16.1-0.20230103093747-1b1febf19315
```
yihuang, line 199: https://github.com/cosmos/iavl/pull/654
Add API `TraverseStateChanges`.

yihuang:
- `versiondb/*.go`, the abstract versiondb implementation, defines the `VersionDB` interface, and implements the cosmos-sdk `MultiStore` and related interfaces.
- `versiondb/tsrocksdb`, the implementation of `VersionDB` interface using the rocksdb user-defined timestamp.
- `versiondb/client`, implements the group of commands: `cronosd changeset` mainly for managing the legacy state migration process.
- `versiondb/extsort`, implements an external sorting algorithm, used in `convert-to-sst` command to sort the data before writing out sst files.
- `versiondb/memiavl`, an in-memory iavl implementation, currently used in `verify` command to replay and verify change set files, in the future, it'll evolves into a [full IAVL implementation](https://github.com/crypto-org-chain/cronos/blob/45162664a0f7bf25f4a3c14db01baf45d737a98a/versiondb/memiavl/README.md).

yihuang:
## `dump` performance

From genesis to `7130689`, 8cores, ssd
```
$ time --verbose cronosd changeset dump --home /chain/.cronosd/ --output data distribution acc authz bank capability cronos evidence evm feegrant feeibc feemarket gov ibc mint params slashing staking transfer upgrade
Percent of CPU this job got: 251%
Elapsed (wall clock) time (h:mm:ss or m:ss): 10:29:45
Maximum resident set size (kbytes): 19423828
$ du -sh data
3.8G	data/
```

## `convert-to-sst `

Convert the plain changeset format to rocksdb sst files with user-defined timestamp feature, spend 19m on testnet node (8cores), the result size: 5.4G, the generated sst files don't overlap with each other (don't trigger further compaction after ingested into versiondb).

The script is like this:
```python
import os
import sys
from datetime import datetime
from pathlib import Path
import subprocess
from concurrent.futures import ThreadPoolExecutor, as_completed

CRONOSD = './result/bin/cronosd'
DEFAULT_STORES = "distribution acc authz bank capability cronos evidence evm feegrant feeibc feemarket gov ibc mint params slashing staking transfer upgrade".split()

SOURCE = Path("./data")
DEST = Path("./sst")

# translate to around 2-3G peak ram usage for each task,
# tune it according to available ram and concurrency level
SORT_CHUNK_SIZE = 256*1024*1024


def run_store(executor, store, srcd, dstd):
    sst = dstd / f'{store}.sst'
    inputs = [src for src in (srcd / store).glob('block-*')]
    return executor.submit(subprocess.run, [CRONOSD, 'changeset', 'convert-to-sst', '--store', store, '--sorter-chunk-size', str(SORT_CHUNK_SIZE), sst] + inputs, check=True)

def main(stores, concurrency):
    DEST.mkdir(exist_ok=True, parents=True)
    with ThreadPoolExecutor(max_workers=concurrency) as executor:
        futs = []
        for store in stores:
            futs.append(run_store(executor, store, SOURCE, DEST))
        for fut in as_completed(futs):
            fut.result()

if __name__ == '__main__':
    try:
        stores = sys.argv[1].split()
    except IndexError:
        stores = DEFAULT_STORES
    main(stores, os.cpu_count())
```

## `ingest-sst`

Ingest the sst files into a rocksdb instance and set latest version, takes less than one second:

```
$ time cronosd changeset ingest-sst /chain/.cronosd/data/versiondb sst/*.sst --move-files --ingest-behind --maximum-version 6698242
```

go.mod
```diff
@@ -192,20 +195,22 @@ replace (
...
 	github.com/evmos/ethermint => github.com/crypto-org-chain/ethermint v0.20.2-cronos
 
 	// Fix upstream GHSA-h395-qcrw-5vmq vulnerability.
 	// TODO Remove it: https://github.com/cosmos/cosmos-sdk/issues/10409
 	github.com/gin-gonic/gin => github.com/gin-gonic/gin v1.7.0
 	github.com/gogo/protobuf => github.com/regen-network/protobuf v1.3.3-alpha.regen.1
-
-	// https://github.com/linxGnu/grocksdb/pull/100
-	github.com/linxGnu/grocksdb => github.com/yihuang/grocksdb v1.7.11-0.20230112025116-f41883a71db3
+	github.com/linxGnu/grocksdb => github.com/yihuang/grocksdb v1.7.11-0.20230120153422-b2f936d8375d
```
JayT106, line 208: didn't see you submit the HyperClockCache changes to the upstream repo, will you submit it or keep it in `crypto-org-chain` repo?

yihuang: Because it needs a pr in rocksdb itself.

versiondb/client/changeset.go
```diff
@@ -0,0 +1,322 @@
...
+	buf := make([]byte, encodedSizeOfKVPair(pair))
+
+	if len(pair.Key) > math.MaxUint16 {
+		return nil, fmt.Errorf("key length %d is too big", len(pair.Key))
+	}
+
+	binary.LittleEndian.PutUint16(buf[1:], uint16(len(pair.Key)))
+	copy(buf[3:], pair.Key)
+	if pair.Delete {
+		buf[0] = 1
```
JayT106, line 124: maybe do return and then remove the else condition?

versiondb/client/changeset.go
```diff
@@ -0,0 +1,322 @@
...
+	reader, err := openChangeSetFile(fileName)
+	if err != nil {
+		return err
+	}
+	defer reader.Close()
+	return fn(reader)
+}
+
+// IterateChangeSets iterate the change set files,
+func IterateChangeSets(
```
JayT106, line 210: this function may take a long time to process, right? Should we add `context` if the user wants to interrupt the process?

versiondb/client/changeset.go
```diff
@@ -0,0 +1,322 @@
...
+) (int64, error) {
+	var (
+		cont               = true
+		lastCompleteOffset int64
+		version, offset    int64
+		changeSet          *iavl.ChangeSet
+		err                error
+	)
+
+	for cont {
```
JayT106, line 222: this for loop may take a long time to process, right? Should we add context if the user wants to interrupt the process?

versiondb/client/changeset.go
```diff
@@ -0,0 +1,322 @@
...
+	fn func(version int64) (bool, error),
+) (int64, error) {
+	var (
+		cont               = true
+		lastCompleteOffset int64
+		version, offset    int64
+		err                error
+	)
+
+	for cont {
```
JayT106, line 257: the same question here.

versiondb/client/changeset.go
```diff
@@ -0,0 +1,322 @@
...
+		nonEmptyFiles []string
+		versions      []uint64
+	)
+
+	for _, fileName := range files {
+		version, err := readVersion(fileName)
+		if err == io.EOF {
+			// skipping empty files
+			continue
+		}
```
JayT106, line 294: should handle other errors like failing to open the file?

go.mod
yihuang: https://github.com/facebook/rocksdb/pull/11110
I'll submit grocksdb PR after rocksdb PR merged.

versiondb/client/changeset.go
yihuang: it's currently only used in cli, user can Ctl+C directly.

versiondb/client/changeset.go
yihuang: good catch, fixed.

versiondb/client/changeset.go
yihuang: it's currently only used in cli, user can Ctl+C directly, I think for enough for now?

versiondb/client/changeset.go
```diff
@@ -0,0 +1,326 @@
...
+	if pair.Delete {
+		buf[0] = 1
+		return buf, nil
+	}
+
+	if len(pair.Value) > math.MaxUint32 {
+		return nil, fmt.Errorf("value length %d is too big", len(pair.Value))
+	}
+	offset := 3 + len(pair.Key)
+	binary.LittleEndian.PutUint32(buf[offset:], uint32(len(pair.Value)))
```
tomtau, line 133: those uint16/uint32 casts should be safe because of those above checks, I assume

versiondb/extsort/sort.go
```diff
@@ -0,0 +1,230 @@
+package extsort
```
tomtau, line 1: is there a reason to have a custom external sort package instead of importing one?

tomtau (commented):
I had a quick look, but didn't review in detail (e.g. I skipped memiavl) -- are there any particular parts you'd like a more detailed review of?

versiondb/client/changeset.go
yihuang: value length int32 should be safe, I also try to push for the `2^16` key length limit in sdk(https://github.com/cosmos/cosmos-sdk/pull/14645), but it seems cosmwasm has a bigger limit.
Maybe we should just use int32 for key as well.

versiondb/extsort/sort.go
yihuang: It's simpler and cleaner than alternatives like https://github.com/lanrat/extsort, and no need to worry about third-party maintenence.
At the beginning I also tried to implement [tornament tree](https://en.wikipedia.org/wiki/K-way_merge_algorithm#Tournament_Tree) as recommended by wikipedia, but give up and just stick with heap in stdlib.

versiondb/client/changeset.go
```diff
@@ -0,0 +1,326 @@
...
+		if err != nil {
+			return err
+		}
+		size += len(buf)
+		items = append(items, buf)
+	}
+
+	var versionHeader [16]byte
+	binary.LittleEndian.PutUint64(versionHeader[:], uint64(version))
+	binary.LittleEndian.PutUint64(versionHeader[8:], uint64(size))
```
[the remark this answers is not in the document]
yihuang: > I assume there are no negative versions?

I think so, even `0` is reserved as special version for genesis in many places.

versiondb/README.md
```diff
@@ -0,0 +1,141 @@
+# VersionDB
```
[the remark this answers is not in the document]
yihuang: Not yet, but some people are interested in our production experience with it.

go.mod
```diff
@@ -192,20 +195,22 @@ replace (
 	// Ref: https://forum.cosmos.network/t/ibc-security-advisory-dragonberry/7702
 	github.com/confio/ics23/go => github.com/confio/ics23/go v0.9.0
 	github.com/cosmos/cosmos-sdk => github.com/cosmos/cosmos-sdk v0.46.7
+
+	// https://github.com/yihuang/iavl/tree/optim_diff_backport
+	github.com/cosmos/iavl => github.com/yihuang/iavl v0.16.1-0.20230111025924-ae78cc50ad97
```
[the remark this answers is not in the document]
yihuang: yes

.github/workflows/build.yml
```diff
@@ -15,6 +15,11 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
+      - uses: cachix/install-nix-action@v18
+        with:
+          nix_path: nixpkgs=channel:nixos-22.11
```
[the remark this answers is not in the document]
yihuang: This is needed to install rocksdb library to build `cronosd`, now librocksdb is always required to build `cronosd`.

versiondb/extsort/sort.go
tomtau: ok, maybe good to document this in the package main's file

versiondb/client/changeset.go
yihuang: I'll update change set format to use varint encoding, and stop worry about length limit here.

versiondb/extsort/sort.go
yihuang: done

versiondb/client/changeset.go
yihuang: done, varint-uint64 encoding now.

yihuang:
> I had a quick look, but didn't review in detail (e.g. I skipped memiavl) -- are there any particular parts you'd like a more detailed review of?

I think correctness-wise, it should be alright, several modules cross-check with each other, for example dump and verify, the versiondb content need to check with integration test, it seems pass all current integration tests fine.
Will need qa team to do more check after devops team did some node migrations.

versiondb/client/dump.go
```diff
@@ -0,0 +1,266 @@
...
+}
+
+type chunk struct {
+	store        string
+	beginVersion int64
+	taskFiles    []string
+	taskGroup    *pond.TaskGroupWithContext
+}
+
+// collect wait for the tasks to complete and concatinate the files into a single output file.
```
calvinaco, line 201:
```suggestion
// collect wait for the tasks to complete and concatenate the files into a single output file.
```

versiondb/client/changeset.go
yihuang: done

versiondb/client/changeset.go
```diff
@@ -0,0 +1,326 @@
...
+	return int64(version), size + 16, &changeSet, nil
+}
+
+// encodedSizeOfKVPair returns the encoded length of a key-value pair
+//
+// layout: deletion(1) + keyLen(2) + key + [ valueLen(4) + value ]
+func encodedSizeOfKVPair(pair iavl.KVPair) int {
+	size := 3 + len(pair.Key)
+	if !pair.Delete {
+		size += 4 + len(pair.Value)
```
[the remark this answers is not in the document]
yihuang: The updated version have much less magic numbers now.

versiondb/client/changeset.go
JayT106: sure.

versiondb/client/dump.go
```diff
@@ -0,0 +1,265 @@
...
+		Use:   "dump store1 store2 ...",
+		Short: "Extract changesets from iavl versions, and save to plain file format",
+		Args:  cobra.MinimumNArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			ctx := server.GetServerContextFromCmd(cmd)
+			if err := ctx.Viper.BindPFlags(cmd.Flags()); err != nil {
+				return err
+			}
+
+			db, err := openReadOnlyDB(ctx.Viper.GetString(flags.FlagHome), server.GetAppDBBackend(ctx.Viper))
```
JayT106, line 41:
```suggestion
			db, err := openReadOnlyDB(ctx.Viper.GetString(flags.FlagHome), server.GetAppDBBackend(ctx.Viper))
			defer db.Close()
```

versiondb/client/dump.go
```diff
@@ -0,0 +1,265 @@
...
+}
+
+func dumpRangeBlocks(outputFile string, tree *iavl.ImmutableTree, blockRange Range) error {
+	fp, err := os.OpenFile(outputFile, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0o600)
+	if err != nil {
+		return err
+	}
+	defer fp.Close()
+
+	writer := snappy.NewBufferedWriter(fp)
```
JayT106, line 183: maybe call defer `writer.Close()` here and then return `nil` at the end?

JayT106: But in line 186, it might just write the changeset into the snappy writer's buffer?

yihuang: yeah, I mean if error happens, it doesn't matter too much if we flush the buffer or not.
And with defer we also better to handle the error `Flush()` returns, it'll be more than a simple line of code, right?

JayT106: yes, we will need to handle the error that happens in `defer`. So in line 186, looks like the writer didn't flush and return directly if I understand correctly, it needs to be flushed before return, right?

yihuang: you mean the callback? it don't need to flush actively, we only need to flush the pending buffer at the very end.

JayT106: Ah, right, it's a callback.

versiondb/extsort/sort.go
```diff
@@ -0,0 +1,235 @@
...
+
+	return NewMultiWayMerge(streams, s.lesserFunc)
+}
+
+// Close closes and remove all the temporary chunk files
+func (s *ExtSorter) Close() error {
+	var err error
+	for _, chunkFile := range s.chunkFiles {
+		err1 := chunkFile.Close()
+		if err != nil {
```
JayT106, line 136: should we append the error instead of assigning it if there is more than one close error?

versiondb/extsort/sort.go
```diff
@@ -0,0 +1,235 @@
...
+	copy(item[:shared], lastItem[:shared])
+	if _, err := io.ReadFull(reader, item[shared:]); err != nil {
+		return nil, err
+	}
+	return item, nil
+}
+
+// diffOffset returns the index of first byte that's different in two bytes slice,
+// the index is encoded as uint8, so the max supported length is `math.MaxUint8`,
+// that should be more than enough in our use case.
```
JayT106, line 220: if we assume max supported length is `math.MaxUint8`. maybe do panic if we see the length exceeds the max?

versiondb/memiavl/persisted_node.go
```diff
@@ -0,0 +1,139 @@
...
+	err2 := blobs.keysFile.Close()
+	err3 := blobs.valuesFile.Close()
+	*blobs = PersistedBlobs{}
+	if err1 != nil {
+		return err1
+	}
+	if err2 != nil {
+		return err2
+	}
+	return err3
```
JayT106, line 61: trying to close 3 files but here might return err1/err2 but other errors happen too.

versiondb/memiavl/persisted_node.go
```diff
@@ -0,0 +1,139 @@
...
+	keyOffset := binary.LittleEndian.Uint64(node.blobs.nodes[node.offset+OffsetKey:])
+	keyLen := uint64(binary.LittleEndian.Uint16(node.blobs.keys[keyOffset:]))
+	keyOffset += 2
+	return node.blobs.keys[keyOffset : keyOffset+keyLen]
+}
+
+// Value result is not defined for non-leaf node.
+func (node PersistedNode) Value() []byte {
+	valueOffset := binary.LittleEndian.Uint64(node.blobs.nodes[node.offset+OffsetValue:])
+	valueLen := uint64(binary.LittleEndian.Uint16(node.blobs.values[valueOffset : valueOffset+2]))
```
JayT106, line 103: doesn't need `+2` like other fields?

versiondb/memiavl/persisted_node.go
```diff
@@ -0,0 +1,139 @@
...
+
+// Right result is not defined for leaf nodes.
+func (node PersistedNode) Right() Node {
+	nodeIndex := binary.LittleEndian.Uint32(node.blobs.nodes[node.offset+OffsetRight:])
+	return PersistedNode{blobs: node.blobs, offset: uint64(nodeIndex) * SizeNode}
+}
+
+func (node PersistedNode) Hash() []byte {
+	offset := node.offset + OffsetHash
+	return node.blobs.nodes[offset : offset+32]
```
JayT106, line 122:
```suggestion
	return node.blobs.nodes[offset : offset+ SizeHash]
```

versiondb/memiavl/snapshot.go
```diff
@@ -0,0 +1,265 @@
...
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+
+	"github.com/ledgerwatch/erigon-lib/mmap"
+)
+
+func WriteSnapshot(root Node, version int64, nodesFile, keysFile, valuesFile, metadataFile string) error {
+	fpNodes, err := os.OpenFile(nodesFile, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0o600)
```
JayT106, line 16: make `os.O_WRONLY|os.O_CREATE|os.O_TRUNC` and `0o600` as const?

versiondb/memiavl/snapshot.go
```diff
@@ -0,0 +1,265 @@
...
+	nodes, nodesHandle, err := Mmap(fpNodes)
+	if err != nil {
+		fpNodes.Close()
+		return nil, 0, 0, err
+	}
+
+	if len(nodes) == 0 {
+		// empty tree
+		if rootIndex != 0 {
+			return nil, 0, 0, errors.New("corrupted snapshot, nodes are empty but rootIndex is not zero")
```
JayT106, line 195: fpNodes.Close()

JayT106, line 195:
```suggestion
                         fpNodes.Close()
			return nil, 0, 0, errors.New("corrupted snapshot, nodes are empty but rootIndex is not zero")
```

versiondb/memiavl/snapshot.go
```diff
@@ -0,0 +1,265 @@
...
+		nodes:  nodes,
+		keys:   keys,
+		values: values,
+
+		nodesHandle:  nodesHandle,
+		keysHandle:   keysHandle,
+		valuesHandle: valuesHandle,
+	}
+
+	return blobs, version, rootIndex, nil
```
JayT106, line 256: any concerns for not using `defer fpNodes.Close()`,  `defer fpKeys.Close()`, `defer fpValues.Close()`?

versiondb/tsrocksdb/comparator.go
```diff
@@ -0,0 +1,52 @@
...
+	case ts1 > ts2:
+		return 1
+	default:
+		return 0
+	}
+}
+
+// compare compares two internal keys with timestamp surfix, larger timestamp comes first.
+func compare(a []byte, b []byte) int {
+	ret := compareWithoutTS(a, true, b, true)
```
JayT106, line 33: do we have other use cases for compareWithoutTS? If no, we can simplify the input to only a & b?

versiondb/tsrocksdb/comparator.go
```diff
@@ -0,0 +1,52 @@
...
+	"bytes"
+	"encoding/binary"
+
+	"github.com/linxGnu/grocksdb"
+)
+
+// CreateTSComparator should be compatible with builtin timestamp comparator.
+func CreateTSComparator() *grocksdb.Comparator {
+	return grocksdb.NewComparatorWithTimestamp(
+		"leveldb.BytewiseComparator.u64ts", TimestampSize, compare, compareTS, compareWithoutTS,
```
JayT106, line 13: should we call different name instead it?

versiondb/tsrocksdb/opts.go
```diff
@@ -0,0 +1,63 @@
...
+	opts.OptimizeLevelStyleCompaction(512 * 1024 * 1024)
+	opts.SetTargetFileSizeMultiplier(2)
+	opts.SetLevelCompactionDynamicLevelBytes(true)
+
+	// block based table options
+	bbto := grocksdb.NewDefaultBlockBasedTableOptions()
+
+	// 1G block cache
+	bbto.SetBlockSize(32 * 1024)
+	bbto.SetBlockCache(grocksdb.NewClockCache(1<<30, 32*1024))
```
JayT106, line 27: Looks like the new rocksdb feature only support after v7.9 , need to update the build requirement in Cronos doc.

versiondb/memiavl/README.md
```diff
@@ -0,0 +1,114 @@
...
+
+- `values`, sequence of length prefixed leaf node values.
+
+  ```
+  size: int32
+  payload
+  ...
+  ```
+
+  Value size is encoded in `int32`, so maximum key length supported is `2**32`, around 4G.
```
JayT106, line 104: should be `2**31 - 1` because not `uint32`?

versiondb/memiavl/README.md
```diff
@@ -0,0 +1,114 @@
...
+
+- `keys`, sequence of length prefixed node keys, deduplicated.
+
+  ```
+  size: int16
+  payload
+  ...
+  ```
+
+  Key size is encoded in `int16`, so maximum key length supported is `65536`.
```
JayT106, line 94: should be `2**15 - 1` if it's `int16`?

versiondb/README.md
```diff
@@ -0,0 +1,141 @@
...
+The legacy state migration process is done in two main steps:
+
+- Extract state change sets from existing archive IAVL tree.
+- Feed the change set files to versiondb.
+
+### Extract Change Sets
+
+```
+$ cronosd changeset dump --home /chain/.cronosd/ --output data distribution acc authz bank capability cronos evidence evm feegrant feeibc feemarket gov ibc mint params slashing staking transfer upgrade
+```
```
JayT106, line 37: for UX, maybe we can add `all` instead of putting these modules?

versiondb/README.md
```diff
@@ -0,0 +1,141 @@
...
+`verify` command will replay all the change sets and rebuild the final IAVL tree and output the root hash of the final version, user can check the root hash against the IAVL tree.
+
+> TODO: will provide command to generate app-hash directly, combining the root hashes of each store, so it'll be easier for end user to verify.
+
+`verify` command takes several minutes and several gigabytes of ram to run, if ram usage is a problem, it can also run incrementally, you can export the snapshot for a middle version, then verify the remaining versions start from that snapshot:
+
+```
+$ cronosd changeset verify --save-snapshot /tmp/snapshot data/acc/block-0.zz data/acc/block-1000000.zz data/acc/block-2000000.zz
+$ cronosd changeset verify --load-snapshot /tmp/snapshot data/acc/block-3000000.zz data/acc/block-4000000.zz data/acc/block-5000000.zz
+```
```
JayT106, line 70: does the op/user requires to verify the changeset in each module manually? meaning verify it just assign the folder path then it will verify all files in the folder for better UX

versiondb/README.md
```diff
@@ -0,0 +1,141 @@
...
+
+#### SST File Writing
+
+To maximize the speed of initial data ingestion into rocksdb, we take advantage of the sst file writer in rocksdb, with that we can write out sst files directly without causing contention on a shared database, the sst files for each store can be written out in parallel. We also developed an external sorting algorithm to sort the data before writing the sst files, so the sst files don't cause extra compaction in the final db.
+
+```
+$ cronosd changeset convert-to-sst --store distribution ./sst/distribution.sst data/distribution/*.zz
+```
+
+`convert-to-sst` will do the sst file writing for a single store, you can wrap it in a simple script to run multiple stores in parallel. Here's an example Python script:
```
JayT106, line 84: maybe put it directly into `scripts` folder. Or just do it in `convert-to-sst` cli directly, for better UX. In any case, we only convert a few modules instead all?

versiondb/README.md
JayT106: The following comments relate to UX and can be improved later I think.

JayT106:
reviewed most of the files except the tests.

versiondb/README.md
yihuang: normally user should convert all modules, but yes, technically it's possible to only convert a subset of modules.

versiondb/README.md
yihuang: depends on if user trust the data provider.

versiondb/README.md
yihuang: I think it's not easy to detect all modules, because we need to include all modules since genesis, if some module is removed in previous upgrade, we can't find it in latest commit info.

versiondb/tsrocksdb/opts.go
yihuang: I was trying to experiment with it, don't have much visible changes, the dump process is still mainly IO bound, I'll convert back to LRU cache first to make this PR simpler.

versiondb/tsrocksdb/comparator.go
yihuang: I deliberate to use the same name as rocksdb builtin comparator, so the result database can still work with builtin tools like `ldb`/`sst_dump`.

versiondb/README.md
tomtau: maybe just a best-effort of hardcoding it statically?

versiondb/memiavl/persisted_node.go
yihuang: fixed

versiondb/extsort/sort.go
yihuang: out of date comment, it use varint encoding, so support int64 technically.

versiondb/extsort/sort.go
yihuang: not sure if it worth the trouble, I'll try to find a library to group errors if there's any.

versiondb/memiavl/snapshot.go
yihuang: added a `createFile` utility to reduce repetition.

versiondb/memiavl/README.md
yihuang: changed to `uint32`.

go.mod
yihuang: don't need the fork anymore, reverted ClockCache for now.

versiondb/README.md
yihuang: > maybe just a best-effort of hardcoding it statically?

that could be misleading to user, if the list out dated, or there maybe difference in different stages of testnet and mainnet, it could be tedious to maintain in the future.

yihuang: I think it better maintained at the doc site, similar to how we document all the breaking upgrades and binary versions.

versiondb/README.md
yihuang: https://github.com/crypto-org-chain/cronos/issues/826
Opened a separate issue to improve cmd UX.

versiondb/README.md
yihuang: https://github.com/crypto-org-chain/cronos/issues/826
Opened a separate issue to improve cmd UX later.

versiondb/README.md
yihuang: https://github.com/crypto-org-chain/cronos/issues/826
Opened a separate issue to improve cmd UX later.

versiondb/extsort/sort.go
yihuang: done, `hashicorp/go-multierror` is good at this

versiondb/memiavl/persisted_node.go
yihuang: done, `hashicorp/go-multierror` is good at this

yihuang:
I removed data block hash index and "OptimizeFiltersForHits" for versiondb, because it don't make as much sense there as the iavl tree.
Also enabled parallel compression, which could have some help on compaction speed.

versiondb/client/convert_to_sst.go
```diff
@@ -0,0 +1,223 @@
...
+
+	aVersion := binary.LittleEndian.Uint64(a[aKeyLen:])
+	bVersion := binary.LittleEndian.Uint64(b[bKeyLen:])
+	// Compare version.
+	// For the same user key with different timestamps, larger (newer) timestamp
+	// comes first.
+	return aVersion > bVersion
+}
+
+func cloneAppend(bz []byte, tail []byte) (res []byte) {
```
mmsqe, line 218: maybe can [cloneAppend](https://github.com/crypto-org-chain/cronos/blob/4900fe0333c7b70c18d18a39fda997e402adb5bc/versiondb/tsrocksdb/store.go#L195) reuse to util, but not sure if need clone since no mutate in both places

versiondb/extsort/delta_encoding.go
```diff
@@ -0,0 +1,89 @@
...
+	nonShared := len(key) - shared
+
+	n1 := binary.PutUvarint(sizeBuf[:], uint64(shared))
+	n2 := binary.PutUvarint(sizeBuf[n1:], uint64(nonShared))
+
+	if _, err := w.Write(sizeBuf[:n1+n2]); err != nil {
+		return err
+	}
+
+	if _, err := w.Write(key[shared:]); err != nil {
```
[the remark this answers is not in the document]
yihuang: yes, delta encoding runs sequentially.

versiondb/client/convert_to_sst.go
yihuang: `append` itself can mutate the first parameter if there's capacity.

versiondb/client/wrapreader.go
```diff
@@ -0,0 +1,51 @@
...
+
+	"github.com/hashicorp/go-multierror"
+)
+
+type wrapReader struct {
+	reader Reader
+	closer io.Closer
+}
+
+// WrapReader wraps reader and closer together to create an new io.ReadCloser.
```
tomtau, line 14:
```suggestion
// WrapReader wraps reader and closer together to create a new io.ReadCloser.
```

versiondb/memiavl/README.md
```diff
@@ -0,0 +1,119 @@
...
+    valueLen: varint-uint64
+    value
+  ]
+
+repeat with next version
+```
+
+- Change set files can be splited with certain block ranges for incremental backup and restoration.
+
+- Historical files can be compressed with zlib, because it don't need to support random access.
```
tomtau, line 57:
```suggestion
- Historical files can be compressed with zlib, because it doesn't need to support random access.
```

versiondb/memiavl/README.md
```diff
@@ -0,0 +1,119 @@
...
+  left    : uint32         // inner node only
+  right   : uint32         // inner node only
+  value   : uint64 offset  // offset in values file, leaf node only
+  hash    : [32]byte
+  ```
+  The node has fixed length, can be indexed directly. The nodes reference each other with the index, nodes are written in post-order, so the root node is always placed at the end.
+
+  Some integers are using `uint32`, should be enough in forseeable future, but could be changed to `uint64` to be safer.
+
+  The implementation will read the mmap-ed content in a zero-copy way, don't use extra node cache, just rely on OS page cache.
```
tomtau, line 88:
```suggestion
  The implementation will read the mmap-ed content in a zero-copy way, won't use extra node cache, it will only rely on the OS page cache.
```

versiondb/memiavl/README.md
```diff
@@ -0,0 +1,119 @@
...
+
+  Value size is encoded in `uint32`, so maximum value length supported is `1<<32-1`, around 4G.
+
+#### Compression
+
+The items in snapshot reference with each other by file offsets, we can apply some block compression techniques to compress keys and values files while maintain random accessbility by uncompressed file offset, for example zstd's experimental seekable format[^1].
+
+### VersionDB
+
+[VersionDB](../README.md) is to support query and iterating historical versions of key-values pairs, currently implemented with rocksdb's experimental user-defined timestamp feature, support query and iterate key-value pairs by version, it's an alternative way to support grpc query service, and much more compact than IAVL tree, similar in size with the compressed change set files.
```
tomtau, line 116:
```suggestion
[VersionDB](../README.md) is to support query and iterating historical versions of key-values pairs, currently implemented with rocksdb's experimental user-defined timestamp feature, support query and iterate key-value pairs by version, it's an alternative way to support grpc query service, and much more compact than IAVL trees, similar in size with the compressed change set files.
```

tomtau (approved):

gomod2nix.toml
```diff
@@ -105,8 +108,9 @@ schema = 3
     version = "v1.0.0"
     hash = "sha256-Qm2aC2vaS8tjtMUbHmlBSagOSqbduEEDwc51qvQaBmA="
   [mod."github.com/cosmos/iavl"]
-    version = "v0.19.4"
-    hash = "sha256-EmpRZ48pjPFq/fIHneut9Vyo5QJATfb3ZO7KzWnqs9g="
+    version = "v0.16.1-0.20230126105431-1bafbada9532"
+    hash = "sha256-eITR5aJLSgSUiVs3mGvrlsydkQIUTyqKl4djXefqn4Y="
+    replaced = "github.com/yihuang/iavl"
   [mod."github.com/cosmos/ibc-go/v5"]
```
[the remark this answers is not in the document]
yihuang: done

versiondb/client/changeset.go
```diff
@@ -0,0 +1,358 @@
...
+	_, err = io.ReadFull(fp, versionBuf[:])
+	if err != nil {
+		return 0, err
+	}
+
+	return binary.LittleEndian.Uint64(versionBuf[:]), nil
+}
+
+// uvarintSize returns the size (in bytes) of `num` encoded as a unsigned varint.
+func uvarintSize(num uint64) int {
```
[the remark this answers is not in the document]
yihuang: done, referenced `binary.PutUvarint` directly.

versiondb/client/convert_to_sst.go
```diff
@@ -0,0 +1,223 @@
...
+//
+// see godoc of `encodeSorterItem` for layout
+func decodeSorterItem(item []byte) ([]byte, iavl.KVPair) {
+	var value []byte
+	keyLen := binary.LittleEndian.Uint32(item[len(item)-SizeKeyLength:])
+	key := item[:keyLen]
+	version := item[keyLen : keyLen+8]
+	delete_ := item[keyLen+8] == 1
+	if !delete_ {
+		value = item[keyLen+9 : len(item)-SizeKeyLength]
```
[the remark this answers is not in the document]
yihuang: yeah, refactored a little bit to reduce magic numbers.

.github/workflows/build.yml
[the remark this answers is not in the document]
yihuang: it's not needed for default `make build` now, but we'll still try both (with or without rocksdb) in CI.

go.mod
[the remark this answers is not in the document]
yihuang: backported in upstream, updated dependency.

yihuang:
it seems can't disable the `typecheck` linter in CI? it don't reproduce locally, will ignore for now.
related: https://github.com/golangci/golangci-lint-action/issues/624

yihuang:
> it seems can't disable the `typecheck` linter in CI? it don't reproduce locally, will ignore for now. related: [golangci/golangci-lint-action#624](https://github.com/golangci/golangci-lint-action/issues/624)

finally figured out, typecheck error just means it fails to compile somewhere, it's CGO dependencies not setup in our case.

PR #795, opened 2022-12-19
Problem: no permissions system in cronos

thomas-nguy:
 
 add the abilities to define admins that has priviledge to execute some admin functions

# PR Checklist:


tomtau (commented):
this is quite a big change -- could there first be a PR that just updates x/cronos's https://github.com/crypto-org-chain/cronos/tree/main/x/cronos/spec , so it's clearer what this is trying to do / what problem this is trying to solve? (I assume the spec won't be up to date if this is added?)

thomas-nguy:
yes, you are totally right. I have created a PR to update the spec there
https://github.com/crypto-org-chain/cronos/pull/801

proto/cronos/query.proto
```diff
@@ -30,6 +30,11 @@ service Query {
         option (google.api.http).get = "/cronos/v1/params";
     }
 
+    // Params queries permissions for a specific address..
+    rpc Permissions(QueryPermissionsRequest) returns (QueryPermissionsResponse) {
+        option (google.api.http).get = "/cronos/v1/permissions";
+    }
```
tomtau, line 36: should this PR bump to "v2", as these extra or breaking changes won't be backported to release/v1.0.x ?

proto/cronos/query.proto
```diff
@@ -86,4 +91,15 @@ message QueryParamsResponse {
...
 
+// QueryPermissionsRequest is the request type for the Query/Permissions RPC method.
+message QueryPermissionsRequest {
+    string address = 1;
+}
+
+// QueryPermissionsResponse is the response type for the Query/Permissions RPC method.
+message QueryPermissionsResponse {
+    bool can_change_token_mapping = 1;
+    bool can_turn_bridge = 2;
```
tomtau, line 102: this is fine, but just for consideration whether this could return a more "extensible" field instead (e.g. a set/vec of permission names, or an integer that would have a bitvector interpretation of permissions)

x/cronos/keeper/msg_server.go
```diff
@@ -64,9 +64,12 @@ func (k msgServer) TransferTokens(goCtx context.Context, msg *types.MsgTransferT
...
 	ctx := sdk.UnwrapSDKContext(goCtx)
-	admin := k.Keeper.GetParams(ctx).CronosAdmin
-	// if admin is empty, no sender could be equal to it
-	if admin != msg.Sender {
+	acc, err := sdk.AccAddressFromBech32(msg.Sender)
+	if err != nil {
+		return nil, err
+	}
+	// check permission
+	if !k.HasPermission(ctx, acc, CanChangeTokenMapping) {
```
tomtau, line 72: maybe ok like this, but was thinking whether it'd make sense to move it to an ante handler, so that all permission-related checking is in one place (I think that SDK's generic circuit breaker will be an ante handler)?

x/cronos/keeper/permissions.go
```diff
@@ -0,0 +1,32 @@
...
+
+const (
+	CanChangeTokenMapping uint64                                  = 1 << iota // 1
+	CanTurnBridge                                                             // 2
+	All                   = CanChangeTokenMapping | CanTurnBridge             // 3
+)
+
+func (k Keeper) SetPermissions(ctx sdk.Context, address sdk.AccAddress, permissions uint64) {
+	store := ctx.KVStore(k.storeKey)
+	permissionsBytes := sdk.Uint64ToBigEndian(permissions)
```
tomtau, line 16: any reason why big endian? or it's a common convention in SDK to store integers like that?

x/cronos/keeper/permissions.go
```diff
@@ -0,0 +1,32 @@
...
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/crypto-org-chain/cronos/x/cronos/types"
+)
+
+const (
+	CanChangeTokenMapping uint64                                  = 1 << iota // 1
+	CanTurnBridge                                                             // 2
+	All                   = CanChangeTokenMapping | CanTurnBridge             // 3
```
tomtau, line 11: I guess the intention here is to have a bitvec for the permissions? If so, maybe good to document that intention to prevent potential mistakes (as if there's one more perm added, it should be 0b100 == 4, and all would be 0b111 == 7)

tomtau (commented):
a few nits for considerations. Should the cronos module and/or protos be bumped to v2?

proto/cronos/query.proto
thomas-nguy: will the next upgrade be v2.0.x or v1.1.x ?

x/cronos/keeper/permissions.go
thomas-nguy: mmm not sure if it is generally adopted convention but ethermint uses this format to store uint64.

proto/cronos/query.proto
tomtau: v2 if it's breaking?

thomas-nguy: yeah that is my assumption but very few cosmos chain (us included) have been following this convention, probably because adding features usually involved breaking changes and want to keep the major version for big breaking changes. 

I'm okay to bump it to v2 (probably in a separate PR) but may need to come to a consensus before hand

cc @yihuang @calvinaco

tomtau: "very few haven't been following this convention" == besides Cronos which just freshly released v1, I can't think of other chains ? cosmos hub was v1 up to v7 now, crypto.org chain was v1 up to v4 now, evmos was v1 up to v10...

thomas-nguy: yeah I've been referring to the module. 

chain-main proto files are still on v1, ethermint and cosmos sdk as well. 

cronos uses the same versioning for both the module and the chain

tomtau: oh ok, I guess proto files can remain v1, but Go mod can be bumped to v2?

x/cronos/keeper/permissions.go
thomas-nguy: added some comments

proto/cronos/query.proto
thomas-nguy: I can do it in a separate PR to keep this one clean

thomas-nguy: https://github.com/crypto-org-chain/cronos/pull/829

mmsqe (commented):
proto-check-breaking is red, you might try `buf mod update`

proto/cronos/query.proto
mmsqe: not sure if we need `proto-swagger-gen` to update swagger as well

thomas-nguy: breakage error is due to some package
```
Failure: buf.build/googleapis/googleapis:80720a488c9a414bb8d4a9f811084989: does not exist
```

I got somehow the same error when I try to generate swagger docs

mmsqe: just tried, seems work after `buf mod update`
```make proto-swagger-gen
Generating Protobuf Swagger
Generate cronos swagger files...
```

thomas-nguy:
```
Generating Protobuf Swagger
Generate cronos swagger files
./cronos/query.proto
Generate cosmos swagger files
./cosmos/app/v1alpha1/query.proto
Failure: buf.build/googleapis/googleapis:80720a488c9a414bb8d4a9f811084989: does not exist
make: *** [proto-swagger-gen] Error 1

```

anyways I managed to generate query.swagger.json under tmp-swagger-gen but seems that a lot of swagger.json is missing from the config

mmsqe: strange, I see this diff after generate
```
modified:   client/docs/statik/statik.go
modified:   client/docs/swagger-ui/swagger.yaml
modified:   third_party/proto/buf.lock
```

thomas-nguy: Do you use `make proto-swagger-gen` ?

Anyways, I'll create a separate ticket to fix the swagger docs.

in config.json I can see its referring to a lot of different path such as 

```
./tmp-swagger-gen/ethermint/evm/v1/query.swagger.json
./tmp-swagger-gen/ethermint/feemarket/v1/query.swagger.json
```

which cannot be found in the repository

since its not related to this PR, we can work on that separately

thomas-nguy: oh nvm it seems that the script combine all of them in the swagger.yaml

```
swagger-combine ./client/docs/config.json -o ./client/docs/swagger-ui/swagger.yaml -f yaml --continueOnConflictingPaths true --includeDefinitions true

```

could you help me push the diff in this PR @mmsqe while I trying to fix my environment ?

x/cronos/keeper/permissions.go
```diff
@@ -0,0 +1,36 @@
...
+
+func (k Keeper) SetPermissions(ctx sdk.Context, address sdk.AccAddress, permissions uint64) {
+	store := ctx.KVStore(k.storeKey)
+	permissionsBytes := sdk.Uint64ToBigEndian(permissions)
+	store.Set(types.AdminToPermissionsKey(address), permissionsBytes)
+}
+
+func (k Keeper) GetPermissions(ctx sdk.Context, address sdk.AccAddress) uint64 {
+	store := ctx.KVStore(k.storeKey)
+	permissionsBytes := store.Get(types.AdminToPermissionsKey(address))
```
mmsqe, line 26: not sure if we should use `store.Has` with addr and type for less gas

thomas-nguy: mmm not sure if it is worth it considering it simplify the design 

also by storing all the permissions within 64bits, we need only one get to fetch all the permissions rather than iterating over all the types for permissions query

proto/cronos/query.proto
thomas-nguy: returning an integer is actually easier but I feel its more readable using this format (especially if its used just for display)

app/ante/ante.go
```diff
@@ -0,0 +1,53 @@
...
+		var anteHandler sdk.AnteHandler
+
+		defer evmante.Recover(ctx.Logger(), &err)
+
+		// Check msg authorization
+		for _, msg := range tx.GetMsgs() {
+			var permissionToCheck uint64
+			var accountToCheck sdk.AccAddress
+
+			switch v := msg.(type) {
```
tomtau, line 25: should this also have MsgUpdatePermissions ?

tomtau (approved):

app/ante/ante.go
thomas-nguy: I was thinking to exclude it because it does not check the permission but only that the sender is equal to admin but could also be added there with `permissionToCheck` = 0

x/cronos/keeper/permissions.go
yihuang: in key it's usually encoded as big endian so it's same order as the number, but for value it should be irrelevant, I'm not sure if golang do optimization for little endian though.

app/ante/ante.go
thomas-nguy: nvm `permissionToCheck = 0` won't work without some refactoring. Since the logic is a bit different I'll leave it out of the ante handler

thomas-nguy:
mmm seems that golangci-lint is failing after the change https://github.com/crypto-org-chain/cronos/commit/7f12ef134453852f0019ba8b0fb7cc143488c512#diff-107e910e9f2ebfb9a741fa10b2aa7100cc1fc4f5f3aca2dfe78b905cbd73c0d2

any clue @yihuang ?

Makefile
```diff
@@ -281,8 +281,8 @@ HTTPS_GIT := https://github.com/crypto-org-chain/cronos.git
 DOCKER := $(shell which docker)
 DOCKER_BUF := $(DOCKER) run --rm -v $(CURDIR):/workspace --workdir /workspace bufbuild/buf
 
-protoVer=v0.7
-protoImageName=tendermintdev/sdk-proto-gen:$(protoVer)
+protoVer=0.12.0
+protoImageName=ghcr.io/cosmos/proto-builder:$(protoVer)
```
thomas-nguy, line 285: required after upgrading go to 1.20

yihuang:
> mmm seems that golangci-lint is failing after the change [7f12ef1#diff-107e910e9f2ebfb9a741fa10b2aa7100cc1fc4f5f3aca2dfe78b905cbd73c0d2](https://github.com/crypto-org-chain/cronos/commit/7f12ef134453852f0019ba8b0fb7cc143488c512#diff-107e910e9f2ebfb9a741fa10b2aa7100cc1fc4f5f3aca2dfe78b905cbd73c0d2)
> 
> any clue @yihuang ?

strange, I can run locally though, there are some errors about sdkerrors deprecation.

Makefile
thomas-nguy: reverted

app/ante/ante.go
```diff
@@ -0,0 +1,53 @@
...
+func NewAnteHandler(options HandlerOptions) (sdk.AnteHandler, error) {
+	return func(
+		ctx sdk.Context, tx sdk.Tx, sim bool,
+	) (newCtx sdk.Context, err error) {
+		var anteHandler sdk.AnteHandler
+
+		defer evmante.Recover(ctx.Logger(), &err)
+
+		// Check msg authorization
+		for _, msg := range tx.GetMsgs() {
```
yihuang, line 21: Normally these logic should be a decorator, but I guess existing api don't support adding extra decorators.

app/ante/ante.go
```diff
@@ -0,0 +1,53 @@
...
+				permissionToCheck = keeper.CanTurnBridge
+				acc, err := sdk.AccAddressFromBech32(v.Sender)
+				if err != nil {
+					panic(err)
+				}
+				accountToCheck = acc
+			}
+
+			if !options.CronosKeeper.HasPermission(ctx, accountToCheck, permissionToCheck) {
+				panic(sdkerrors.Wrap(sdkerrors.ErrInvalidAddress, "msg sender is unauthorized"))
```
yihuang, line 43: why not return error?

yihuang (commented):
why not the ante handler return error instead of panic?
do we still need the CronosAdmin in parameters?

app/ante/ante.go
thomas-nguy: changed

app/ante/ante.go
thomas-nguy: its a bit a hacky way

yihuang (approved):

yihuang:
> CronosAdmin is still needed for the sake of decentralisation. It is defined by governance and only the CronosAdmin can add/delete/update permissions

oh, i thought the set permissios stuff is supposed to be controlled by gov

thomas-nguy:
its controlled by the new MsgUpdatePermissions that only admin can use

PR #801, opened 2023-01-04
Problem: No record for ADR-009 permissioned addresses in Cronos

thomas-nguy:

# PR Checklist:


docs/architecture/adr-009.md
```diff
@@ -0,0 +1,62 @@
...
+
+- Keep the admin address in Cronos param's space. The admin has all the permissions and has the ability to change the permission of an address. 
+
+- By default, all addresses in Cronos are non-permissioned
+
+- Assign to each "restricted" messages in Cronos a permission (integer value) and create in Cronos module a mapping between addresses and permissions that is stored in memory. For now, there are only two messages that require permission : MsgUpdateTokenMapping and MsgTurnBridge.
+
+- Create a msg type "MsgUpdatePermissions" that only admin can use and allow to update the address permission mapping.
+
+- Change the logic to always check for the permission before processing the restricted messages.
```
tomtau, line 42: FYI, all/most of this functionality (admin policies to maintain a group membership, having roles/permissions for different addresses based on policies for different actions, etc.)  is pretty much implemented in the x/group module: https://github.com/cosmos/cosmos-sdk/tree/main/x/group#concepts

Regarding the decision, my suggestion would be:

* move away from x/params, as it's deprecated, and migrate/store those top-level admin role fields in the module
* see if x/group can be leveraged (as it's a kind of "multi-sig+permissions on steroids"):
1. either directly (by having MsgUpdateTokenMapping and MsgTurnBridge controlled by a specific x/group)
2. or if not possible, by re-using some of x/group's logic: yesterday, we had a call with @tac0turtle and he mentioned there was a proposal to modularise x/group to make it composable, such that non-SDK modules can import its logic and use it internally. He's not sure whether he'd find time for it, but if this is needed, perhaps @adu-crypto can take a stab at it?

thomas-nguy: - Oh right, admin is probably already moved from x/params by latest @adu-crypto changes. I need to update the ADR

-  I have looked into x/group but it seems it only allows to define policies and would requires a certain number of vote for the policy to pass. It is quite different from the use case presented, at least for the circuit breaker which requires permissioned addresses to have 100% power over an action (one or many could have the same priviledges)

I need to check more what could be leveraged from x/group but the implementation in https://github.com/crypto-org-chain/cronos/pull/795 is actually very simple and requires few change in the code.

tomtau: > requires permissioned addresses to have 100% power over an action (one or many could have the same priviledges)

that can just be set with a policy within the group.

> Is it still worth to try to make use of the module?

That depends on whether this anticipates there'll be a need for more control-related messages (e.g. with future gravity bridge changes) or more flexibility in the specifying pre-conditions for message execution.

BTW, can `MsgTurnBridge` be implemented using `MsgTripCircuitBreaker` and specifying the bridge message type urls https://github.com/cosmos/cosmos-sdk/pull/14229/files#diff-a6027a5f667c19d32fd0bcb47ca910d4e1f9ce10d02a6f8eda08e0109a32dd19R15 ?

thomas-nguy:
```
that can just be set with a policy within the group.
```

Right, but it makes the operation step a bit overkill. Anyways I am open to all form of suggestions.

```
BTW, can MsgTurnBridge be implemented using MsgTripCircuitBreaker and specifying the bridge message type urls https://github.com/cosmos/cosmos-sdk/pull/14229/files#diff-a6027a5f667c19d32fd0bcb47ca910d4e1f9ce10d02a6f8eda08e0109a32dd19R15 ?
```

I think if we want to achieve the same effect, then it is possible to pause all msgs related to the bridge (MsgSubmitEthereumEvent, MsgSendToEthereum, MsgCancelSendToEthereum, MsgRequestBatchTx....)

But we wont be able to stop the logic in the abci, consequently batches will continue to be generated as long as there are pending unbatched transactions in the pool. Depending if the attacker has the ability to sign the batch alone, there is still a risk of attack.

thomas-nguy: I have edited the ADR to include x/group module option

docs/architecture/adr-009.md
```diff
@@ -0,0 +1,88 @@
...
+
+
+### Option 2 : Use a subset of x/group module functionality to restrict some transactions in cronos
+
+For the above problem, we propose the following:
+
+- For each specific msg, we define a group and for the group, a group policy
+- Use the admin define in cronos space as general admin for the group
+- Set the policy of the group to ``threshold: "1"`` , ``min_execution_period: 0s`` and ``voting_period: 86400s``
+- The policy is associated to a group policy address, each permissioned msg within the proposal needs to be signed by the group policy address (TBD)
```
thomas-nguy, line 63: not sure how to get access to the group policy address signature. Perhaps with x/authz. Need further investigations

docs/architecture/adr-009.md
```diff
@@ -0,0 +1,89 @@
...
+- To be abe to execute a permissioned msg, one member of the group needs to send the msg ``Msg/SubmitProposal``
+
+In Cronos module, we need to restrict ``MsgUpdateTokenMapping`` and ``MsgTurnBridge`` msg to only specific "group policy address".
+
+- Store in the state a mapping between msg and group policy address that are allowed to trigger the msg
+- Only a governance proposal to be able to change the mapping
+- Add code to check the sender address of the message and reject it if it does not belong to one of the address defined in the mapping
+
```
thomas-nguy, line 71: It seems the change in x/Cronos is a bit similar with what we need with proposal 1. Not sure if it can be simplified

docs/architecture/adr-009.md
```diff
@@ -0,0 +1,89 @@
...
+- Admin of the group cannot be changed through governance.
+
+
+## Status
+
+Proposed
+
+## References
+
+- https://github.com/crypto-org-chain/cronos/pull/795
```
thomas-nguy, line 89:
```suggestion
- https://github.com/crypto-org-chain/cronos/pull/795
- https://github.com/cosmos/cosmos-sdk/tree/main/x/group#xgroup
```

docs/architecture/adr-009.md
```diff
@@ -0,0 +1,90 @@
...
+- Require some change in the code.
+- Not flexible for potential future change which may require functionalities on the group module.
+
+
+### Option 2 : Use a subset of x/group module functionality to restrict some transactions in cronos
+
+For the above problem, we propose the following:
+
+- For each specific msg, we define a group and for the group, a group policy
+- Use the admin define in cronos space as general admin for the group
```
adu-crypto, line 61:
```suggestion
- Use the admin defined in cronos space as general admin for the group
```

adu-crypto (commented):
I'm wondering if we could monitor those metrics on-chain and ask for execution of specific message automatically given the `x/group` module.
the option 1 seems simple and flexible because we could do most of the monitoring work off-chain and only add the permissions map to the cronos store, while we need to specify the policy if we choose option 2.

thomas-nguy:
I havent found anything related to "restrict message to a certain group" in x/module. Maybe in x/authz? but it seems more related to module's authorization.

docs/architecture/adr-009.md
tomtau: > I think if we want to achieve the same effect, then it is possible to pause all msgs related to the bridge (MsgSubmitEthereumEvent, MsgSendToEthereum, MsgCancelSendToEthereum, MsgRequestBatchTx....)
> But we wont be able to stop the logic in the abci, consequently batches will continue to be generated as long as there are pending unbatched transactions in the pool. Depending if the attacker has the ability to sign the batch alone, there is still a risk of attack.

I checked in the cosmos tech channel and while this functionality (a way for modules to pause begin/end blocker functionality) isn't in the initial SDK's circuit breaker (the initial version will only have the circuit breaker admin permissions logic + pausing messages), there was a positive feedback that something like this could be included afterwards in SDK.
Given cronos already includes x/crisis, it's probably good to leverage the SDK's generic circuit breaker instead of duplicating it -- so the decision record could describe either it'd be used or how the migration could be done if there wass a custom duplicate one.

thomas-nguy: I have added a section regarding the wip circuit breaker.

While the implementation is mostly related to x/gravity rather than x/cronos, I have kept the details out of the ADR and created a ticket in the gravity bridge repository to investigate/work on some potential replacement https://github.com/crypto-org-chain/gravity-bridge/issues/194

docs/architecture/adr-009.md
```diff
@@ -0,0 +1,99 @@
...
+An address in Cronos that carry no permissions. 
+
+
+## Decision
+
+While there is some work-in-progress from Cosmos-sdk team to design a circuit breaker module (https://github.com/cosmos/cosmos-sdk/issues/14226) that could potentially resolve our issue, a temporary solution is to implement our own admin system in cronos for certain message type.
+
+After the circuit breaker is made available, those solutions could be either:
+- completely remove (rollback) and be replaced by the circuit breaker
+- partially removed and be replaced by the circuit breaker
```
thomas-nguy, line 36: for` MsgUpdateTokenMapping` it is not related to circuit breaker. We might want to keep this solution for this msg and use the circuit breaker for `MsgTurnBridge`

docs/architecture/adr-009.md
```diff
@@ -0,0 +1,99 @@
...
+- work along with the circuit breaker
```
thomas-nguy, line 37: in case the design of the circuit breaker does not fit our requirements

docs/architecture/adr-009.md
thomas-nguy: after checking, no need for the signature but all message must originate from group policy address (typically the field from)

thomas-nguy:
@tomtau @calvinaco @yihuang 

I have updated the ADR with final decision. Let me know if it is okay to merge

docs/architecture/adr-009.md
```diff
@@ -0,0 +1,104 @@
+# ADR 009: Permissioned addresses in Cronos
+
+## Changelog
+* 2022-01-04: first draft
+
+## Context
+
+Some messages in cronos require permissions. For example changing the mapping to support new CRC20 auto-conversion contracts or disabling the bridge module in case of emergency. Right now, only the admin has the ability to use those messages.
+The admin is a single address defined in cronos space and can be modified through governance. It is usually a multi-sig address shared by multiple trusted party in order to achieve a higher level of security.
```
tomtau, line 9:
```suggestion
Some messages in cronos require permissions. For example, changing the mapping to support new CRC20 auto-conversion contracts or disabling the bridge module in case of emergency. Right now, only the admin has the ability to use those messages.
 The admin is a single address defined in cronos space and can be modified through governance. It is usually a multi-sig address shared by multiple trusted parties in order to achieve a higher level of security.
```

docs/architecture/adr-009.md
```diff
@@ -0,0 +1,104 @@
...
+
+While a single multi-sig admin address was originally implemented with simplicity in mind, realistically it is not practical to rely on a single address to perform all admin tasks.
+As those operations could become more frequent (updating the token mapping) or needs to be trigger by external systems (circuit breaker for gravity module), it would be more practical to define a granular permission system which restrict certain operations to only "some" known addresses.
```
tomtau, line 12:
```suggestion
As those operations could become more frequent (updating the token mapping) or needs to be triggered by external systems (circuit breaker for gravity module), it would be more practical to define a granular permission system which restricts certain operations to only some known addresses.
```

tomtau (approved):

calvinaco (approved):

yihuang (approved):

PR #946, opened 2023-03-31
Problem: memiavl changes are not persisted in real time

pysel:
Closes: #947 

# PR Checklist:


memiavl/wal.go
```diff
@@ -0,0 +1,124 @@
...
+	return Change{
+		Delete: value == nil,
+		Key:    key,
+		Value:  value,
+	}
+}
+
+// writeChange writes a change to the write-ahead log.
+// Returns bytes written
+func writeChange(wal *wal.Log, change Change, index uint64) (uint64, error) {
```
yihuang, line 67: I suggest storing the change set of a whole block as a single entry in wal, since that's the atomic unit to replay.

yihuang:
https://github.com/crypto-org-chain/cronos/issues/947 issue created

memiavl/wal.go
pysel: makes sense! thx

yihuang: For simplicity we can reuse the [change set formats used in versiondb](https://github.com/crypto-org-chain/cronos/blob/bd7251b866bbe3944b997ea002ab846a0dccf6cb/versiondb/client/changeset.go#L31), in memory type would be `[]iavl.KVPair`

yihuang: https://github.com/crypto-org-chain/cronos/pull/952

You can build on top of this API, I've implemented the marshal/unmarshal for the `ChangeSet` type.

yihuang: Another thing is someone need to receive the raw `Set`/`Remove` calls and accumulate the change set, I think we can let the `KVStore` wrapper to do that, so memiavl only expose the `ApplyChangeSet` API, rather than the raw `Set/Remove` APIs.

pysel: oh, I already implemented it but yeah no problem I'll merge your branch, thx for letting me know!

yihuang: I just did a little bit more changes, so now memiavl tree only expose a single method `ApplyChangeSet` to do modifications, I think you can merge the branch now, I just need some review from colleagues to merge.

yihuang: Just realize there's `iavl.ChangeSet` type already, so I changed the PR to reuse that type.

yihuang: #952 is merged.

pysel: applied those changes to this PR

memiavl/tree.go
```diff
@@ -68,10 +81,12 @@ func (t *Tree) ApplyChangeSet(changeSet *iavl.ChangeSet, updateHash bool) ([]byt
 }
 
 func (t *Tree) set(key, value []byte) {
+	t.bwal.addChange(preparePair(key, value))
```
yihuang, line 84: now you can add the whole change set in `ApplyChangeSet` at once?

pysel: yeah, I refactored the code today to use the changeset PR's logic. So basically wal works right now like this: it stores a pair in memory when a `set/remove` is called, and writes to the underlying wal every block's changeset when a `saveVersion` is called.

Right now I am working on `ReplayWAL` function that will basically take the tree at version X, and replay all the changesets stored in a wal until the desired height.

If you have any suggestions on the design, please, lmk!

yihuang: I think you can write the change set passed in ApplyChangeSet method directly into WAL, since that's the only entry method now, set/remove won't be called directly by user.

pysel: yeah, makes sense, will do that

memiavl/tree.go
```diff
@@ -16,54 +18,70 @@ type Tree struct {
...
 func (t *Tree) ApplyChangeSet(changeSet *iavl.ChangeSet, updateHash bool) ([]byte, int64, error) {
+	t.bwal.BlockChangeset = *changeSet
 	for _, pair := range changeSet.Pairs {
 		if pair.Delete {
 			t.remove(pair.Key)
 		} else {
 			t.set(pair.Key, pair.Value)
 		}
 	}
+	t.bwal.Flush()
```
yihuang, line 84: Should we just call `t.bwal.writeBlockChanges ` directly here?

memiavl/wal.go
```diff
@@ -0,0 +1,143 @@
...
+	return blockWAL{
+		wal:            log,
+		version:        version,
+		BlockChangeset: iavl.ChangeSet{},
+	}, nil
+}
+
+// prepareChange prepares a change for write-ahead log.
+// If value provided is nil, it implies the Change should indicate deletion.
+func preparePair(key []byte, value []byte) iavl.KVPair {
```
yihuang, line 55: there seems some dead codes.

memiavl/tree.go
pysel: `Flush()` does that internally. but I see what you mean. I think I can just get rid of `BlockChangeset` in wal struct, because we do not need it anymore, since `set/remove` are not exposed

memiavl/tree.go
```diff
@@ -121,3 +139,41 @@ func (t *Tree) Get(key []byte) []byte {
...
+		if err != nil {
+			return fmt.Errorf("failed to read changeset with index %d from wal: %w", i, err)
+		}
+
+		blockChanges, err := UnmarshalChangeSet(bz)
+		if err != nil {
+			return err
+		}
+
+		changesets = append(changesets, *blockChanges)
```
yihuang, line 167: - I think we can call `ApplyChangeSet` here directly, so don't need to decode all the change sets in memory.
- we don't have to update hash for the intermediate versions, we can do `tree.ApplyChangeSet(, false)`, and only do a `tree.RootHash()` at the end.

memiavl/tree.go
yihuang: yeah, that'd be simpler.

pysel: ok, changed that, now changesets can be directly provided as `Flush` args

pysel: I will also move `Flush` to be ran before all sets and removes

memiavl/tree.go
pysel: 1) If we do that, if a user provides a version that does not exist in wal, the changes to the tree will be applied up to the version that exists in wal and then error. Is that desired?

pysel: I also will add an option to simply sync all versions available in the wal

pysel: added a way to sync untill the end of wal: if `untilVersion` is 0, it will sync till the end

pysel: about the error still seems weird to me, it would mean that the tree is corrupted, but why would we declare it corrupted if that can be prevented? But the final decision is up to you

pysel: also want to add function to read batches of changesets from wal in one go, will improve readability in `ReplayWAL` I believe

pysel: nvm, I am using the native `wal.Read()` in `ReplayWAL`, not the wrapper, but I will propose batch reads to tidwall/wal

yihuang: > about the error still seems weird to me, it would mean that the tree is corrupted, but why would we declare it corrupted if that can be prevented? But the final decision is up to you

just want to lower the memory usage here, it could be pretty big depending on the snapshot interval.

yihuang (approved):
LGTM, thanks a lot for the contribution, I'll probably do some adjustments when doing the final integration.

yihuang:
@pysel can you run `go mod tidy && gomod2nix` in root directory, or allow me to push the changes to the branch?

memiavl/tree.go
pysel: yeah your point makes sense. What do you think about copying the tree at the beginning of the function and applying changes to it the way you suggested? still a memory overhead, but not that big I guess

yihuang: I think if anything unexpected happens with WAL, it's almost always a serious unrecoverable issue.

pysel: oh, I see what you mean. What I meant here is a user's mistake, aka providing an `untilVersion` that is higher than the latest version from the wal, which could result in `wal.Read()` error. I will just add a check that the version provided is within boundaries and we should be good to go with your design, thanks!

yihuang (changes requested):
The wal Flush seems always fail for "out of order" error

memiavl/tree.go
```diff
@@ -16,47 +18,82 @@ type Tree struct {
...
 		root:    snapshot.RootNode(),
-	}
+		bwal:    wal,
+	}, nil
 }
 
 // ApplyChangeSet apply the change set of a whole version, and update hashes.
-func (t *Tree) ApplyChangeSet(changeSet *iavl.ChangeSet, updateHash bool) ([]byte, int64, error) {
+// Returns hash, new version, and potential error.
+func (t *Tree) ApplyChangeSet(changeSet iavl.ChangeSet, updateHash bool) ([]byte, int64, error) {
```
mmsqe, line 90: yes, pointer was wrong [use](https://medium.com/swlh/use-pointer-of-for-range-loop-variable-in-go-3d3481f7ffc9), that var will be reused in for loop

pysel:
> The wal Flush seems always fail for "out of order" error in unit test.

checking

pysel:
@yihuang ah, the reason is that the wal file is not deleted at the end of some tests, so it tries to write indices to the file that already exist there since the last version, let me fix it

yihuang (approved):

memiavl/tree_test.go
```diff
@@ -107,6 +118,133 @@ func TestNewKey(t *testing.T) {
...
+		"random wal path: will create empty new wal: should fail": {
+			untilVersion: 0,
+			tree:         tree2,
+			walPath:      randomWalPath,
+			expectingErr: true,
+		},
+	}
+
+	for name, tc := range tests {
+		t.Run(name, func(t *testing.T) {
```
mmsqe, line 207: might need mv the Cleanup before each test

pysel: these subtests do not create any new wals, so I do not think we should

mmsqe (approved):

PR #964, opened 2023-04-11
Problem: old snapshots are not pruned

mmsqe:

# PR Checklist:

### DIFF
--- a/memiavl/db.go
+++ b/memiavl/db.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/pkg/errors"
 	"github.com/tidwall/wal"
@@ -97,13 +98,28 @@ func (db *DB) Commit(changeSets MultiChangeSet) ([]byte, int64, error) {
 			}
 
 			// snapshot rewrite succeeded, catchup and switch
-			// TODO prune the old snapshots
 			if err := result.mtree.CatchupWAL(db.wal); err != nil {
 				return nil, 0, fmt.Errorf("catchup failed: %w", err)
 			}
 			if err := db.reloadMultiTree(result.mtree); err != nil {
 				return nil, 0, fmt.Errorf("switch multitree failed: %w", err)
 			}
+			// prune the old snapshots
+			version := uint32(db.lastCommitInfo.Version)
+			latestSnapshot := snapshotName(version)
+			prefix := strings.TrimSuffix(latestSnapshot, fmt.Sprintf("%d", version))
+			filepath.WalkDir(db.dir, func(path string, entry os.DirEntry, err error) error {
+				if err != nil {
+					return err
+				}
+
+				if entry.IsDir() &&
+					strings.HasPrefix(entry.Name(), prefix) &&
+					entry.Name() != latestSnapshot {
+					os.RemoveAll(path)
+				}
+				return nil
+			})
 		default:
 		}
 	}
@@ -213,8 +229,12 @@ func (db *DB) Close() error {
 	return stderrors.Join(db.MultiTree.Close(), db.wal.Close())
 }
 
+func snapshotName(version uint32) string {
+	return fmt.Sprintf("snapshot-%d", version)
+}
+
 func snapshotPath(root string, version uint32) string {
-	return filepath.Join(root, fmt.Sprintf("snapshot-%d", version))
+	return filepath.Join(root, snapshotName(version))
 }
 
 func currentPath(root string) string {


memiavl/db.go
```diff
@@ -97,13 +98,28 @@ func (db *DB) Commit(changeSets MultiChangeSet) ([]byte, int64, error) {
...
 				return nil, 0, fmt.Errorf("catchup failed: %w", err)
 			}
 			if err := db.reloadMultiTree(result.mtree); err != nil {
 				return nil, 0, fmt.Errorf("switch multitree failed: %w", err)
 			}
+			// prune the old snapshots
+			version := uint32(db.lastCommitInfo.Version)
+			latestSnapshot := snapshotName(version)
+			prefix := strings.TrimSuffix(latestSnapshot, fmt.Sprintf("%d", version))
+			filepath.WalkDir(db.dir, func(path string, entry os.DirEntry, err error) error {
```
yihuang, line 111: `filepath.WalkDir` is recursive, `os.ReadDir` is better here?

memiavl/db.go
```diff
@@ -97,13 +98,28 @@ func (db *DB) Commit(changeSets MultiChangeSet) ([]byte, int64, error) {
...
 			if err := result.mtree.CatchupWAL(db.wal); err != nil {
 				return nil, 0, fmt.Errorf("catchup failed: %w", err)
 			}
 			if err := db.reloadMultiTree(result.mtree); err != nil {
 				return nil, 0, fmt.Errorf("switch multitree failed: %w", err)
 			}
+			// prune the old snapshots
+			version := uint32(db.lastCommitInfo.Version)
+			latestSnapshot := snapshotName(version)
+			prefix := strings.TrimSuffix(latestSnapshot, fmt.Sprintf("%d", version))
```
yihuang, line 110: The prefix should be just a constant, "snapshot-"

yihuang (changes requested):
We better make the pruning strategy configurable, the basic strategy is how many snapshots to keep, the old snapshots are useful sometimes to recover old version of trees, especially with the help of WAL.

memiavl/db.go
mmsqe: it's const after trim, do you mean hardcode to `snapshot-`?

yihuang: yes

memiavl/db.go
```diff
@@ -97,13 +100,24 @@ func (db *DB) Commit(changeSets MultiChangeSet) ([]byte, int64, error) {
...
-			// TODO prune the old snapshots
 			if err := result.mtree.CatchupWAL(db.wal); err != nil {
 				return nil, 0, fmt.Errorf("catchup failed: %w", err)
 			}
 			if err := db.reloadMultiTree(result.mtree); err != nil {
 				return nil, 0, fmt.Errorf("switch multitree failed: %w", err)
 			}
+			// prune the old snapshots
+			version := uint32(db.lastCommitInfo.Version)
+			latestSnapshot := snapshotName(version)
```
yihuang, line 111:
```suggestion
			latestSnapshot := snapshotName(db.snapshot.Version())
```

The latest snapshot should be the loaded snapshot version, rather than the latest version.

yihuang, line 111:
```suggestion
			latestSnapshot := snapshotName(db.snapshot.Version())
```

The latest snapshot should be the loaded snapshot version, rather than the latest version.

memiavl/db.go
```diff
@@ -97,13 +100,24 @@ func (db *DB) Commit(changeSets MultiChangeSet) ([]byte, int64, error) {
...
+			entries, err := os.ReadDir(db.dir)
+			if err == nil {
+				for _, entry := range entries {
+					if entry.IsDir() && strings.HasPrefix(entry.Name(), SnapshotPrefix) &&
+						entry.Name() != latestSnapshot {
+						os.RemoveAll(filepath.Join(db.dir, entry.Name()))
```
yihuang, line 117: Handle the error return?

memiavl/db.go
mmsqe: we use lastCommitInfo.Version [here](https://github.com/crypto-org-chain/cronos/blob/ae728e2f401f57a87b7ce09db04683c1208e932c/memiavl/db.go#L154), loaded snapshot version is lastCommitInfo when write latest?

memiavl/db.go
mmsqe: I think can ignore the error here actually

memiavl/db.go
yihuang: `CatchupWAL` will move the in-memory tree forward

memiavl/db.go
```diff
@@ -217,7 +220,7 @@ func (db *DB) RewriteSnapshotBackground() error {
 			return
 		}
 
-		ch <- snapshotResult{mtree: mtree}
+		ch <- snapshotResult{mtree: mtree, path: path}
```
yihuang, line 223: An alternative is simply return the snapshot version here.

mmsqe: done, but I think file path seems more related

memiavl/db.go
```diff
@@ -204,7 +209,7 @@ func (db *DB) RewriteSnapshotBackground() error {
 	wal := db.wal
 	go func() {
 		defer close(ch)
-		path, err := cloned.RewriteSnapshot()
+		version, err := cloned.RewriteSnapshot()
```
yihuang, line 212: `version` is just the `cloned.lastCommitInfo.Version`?

mmsqe: it's return from RewriteSnapshot, CatchupWAL will update it

yihuang: The CatchupWAL is done on the newly loaded `mtree`, the `cloned` db is not updated.

memiavl/db.go
```diff
@@ -97,13 +100,27 @@ func (db *DB) Commit(changeSets MultiChangeSet) ([]byte, int64, error) {
...
 
 			// snapshot rewrite succeeded, catchup and switch
-			// TODO prune the old snapshots
 			if err := result.mtree.CatchupWAL(db.wal); err != nil {
 				return nil, 0, fmt.Errorf("catchup failed: %w", err)
 			}
 			if err := db.reloadMultiTree(result.mtree); err != nil {
 				return nil, 0, fmt.Errorf("switch multitree failed: %w", err)
 			}
+			// prune the old snapshots
```
yihuang, line 109: do the pruning in a background goroutine?

memiavl/db.go
yihuang: make sense, in the future we should add an error log then.
And I think it makes more sense to do the pruning in a background go-routine.

memiavl/db.go
```diff
@@ -111,15 +111,16 @@ func (db *DB) Commit(changeSets MultiChangeSet) ([]byte, int64, error) {
 			latestSnapshot := snapshotName(version)
 			entries, err := os.ReadDir(db.dir)
 			if err == nil {
-				for _, entry := range entries {
-					if entry.IsDir() && strings.HasPrefix(entry.Name(), SnapshotPrefix) &&
-						entry.Name() != latestSnapshot {
-						err := os.RemoveAll(filepath.Join(db.dir, entry.Name()))
-						if err != nil {
-							fmt.Printf("failed when remove old snapshot: %s\n", err)
+				go func() {
```
yihuang, line 114: how about include the `os.ReadDir` as well, that's also IO if not cached.

memiavl/db.go
```diff
@@ -171,8 +190,9 @@ func (db *DB) reloadMultiTree(mtree *MultiTree) error {
 }
 
 type snapshotResult struct {
-	mtree *MultiTree
-	err   error
+	mtree   *MultiTree
+	err     error
+	version uint32
```
yihuang, line 195: rename `version` to sth like `snapshotVersion`, to distinguish with latest version of `mtree` which has replay WAL.

memiavl/db.go
```diff
@@ -188,7 +205,8 @@ func (db *DB) RewriteSnapshotBackground() error {
 	wal := db.wal
 	go func() {
 		defer close(ch)
-		if err := cloned.RewriteSnapshot(); err != nil {
+		err := cloned.RewriteSnapshot()
+		if err != nil {
```
yihuang, line 209:
```suggestion
		if err := cloned.RewriteSnapshot(); err != nil {
```

yihuang (approved):

PR #1004, opened 2023-04-25
Problem: wal write is heavy in block replaying

yihuang:
Solution:
    - introduce `store.async-wal` option to write wal asynchronously
    - rename the `memiavl` flag to `store.memiavl`.

In the block replaying test, it reduce the sync time from 2m to 1m10s, 40% reduction.

# PR Checklist:

### DIFF
--- a/memiavl/db.go
+++ b/memiavl/db.go
@@ -172,7 +172,7 @@ func (db *DB) ApplyUpgrades(upgrades []*TreeNameUpgrade) error {
 func (db *DB) checkAsyncTasks() error {
 	return errors.Join(
 		db.checkAsyncWAL(),
-		db.cleanupSnapshotRewrite(),
+		db.checkBackgroundSnapshotRewrite(),
 	)
 }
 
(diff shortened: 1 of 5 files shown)


memiavl/db.go
```diff
@@ -142,29 +168,44 @@ func (db *DB) ApplyUpgrades(upgrades []*TreeNameUpgrade) error {
...
+		db.checkBackgroundSnapshotRewrite(),
+	)
+}
+
+// checkAsyncWAL check the quit signal of async wal writing
+func (db *DB) checkAsyncWAL() error {
+	select {
+	case err := <-db.walQuit:
+		// async wal writing failed, we need to abort the state machine
+		return fmt.Errorf("async wal writing goroutine quit unexpectedly: %w", err)
```
mmsqe, line 184: need handle nil here
```suggestion
		if err != nil {
			// async WAL writing failed, we need to abort the state machine
			return fmt.Errorf("async WAL writing goroutine quit unexpectedly: %w", err)
		}
```

mmsqe (commented):
so async is faster than batch?

memiavl/db.go
yihuang: it also treat `nil` as an error here, the async wal goroutine just not expected to quit.

yihuang:
> so async is faster than batch?

yes, batch don't improve much, async improve a lot.

memiavl/db.go
mmsqe: close will return right?

yihuang: yes, but whenever it quit, the state machine should be aborted

PR #1014, opened 2023-04-28
Problem: no stateful precompiled contract for relayer

mmsqe:
Updates: https://github.com/cosmos/relayer/issues/1187

# PR Checklist:


integration_tests/ibc_utils.py
```diff
@@ -16,9 +17,10 @@ class IBCNetwork(NamedTuple):
     chainmain: Chainmain
     hermes: Hermes
     incentivized: bool
+    proc: subprocess.Popen[bytes] | None
 
 
-def prepare_network(tmp_path, file, incentivized=True, start_relay=True):
+def prepare_network(tmp_path, file, incentivized=True, is_relay=True, is_hermes=True):
```
yihuang, line 23: I guess it can't do both, so maybe `relayer = 'hermes | gorelayer'` is better?

thomas-nguy:
I guess this PR is to be able to test the ibc middleware that will convert cosmos msgs to ethermint msgs?

would it be possible to write an ADR for the design? @yihuang

yihuang:
I guess this PR itself is just try go-relayer in integration test, so it'll be more convenient for @mmsqe to further experiment using the golang relayer

thomas-nguy:
yes I understand that this PR is only to integrate with the go-relayer but I just using this opportunity to raise the need for an ADR before getting too deep in the implementation so that we can have a better visibility on what has to be done

thomas-nguy:
Ideally the IBC protocol should be relayer agnostic so using hermes or gorelayer unless we change the semantic for cronos which may raise some concerns (which could be discussed on the ADR)

yihuang:
> Ideally the IBC protocol is relayer agnostic so using hermes or gorelayer implementation should not matter unless we change the semantic for cronos which may raise some concerns (which could be discussed in the ADR)

@mmsqe need to patch the relayer to let it send evm tx instead of cosmos tx.

mmsqe:
> > Ideally the IBC protocol is relayer agnostic so using hermes or gorelayer implementation should not matter unless we change the semantic for cronos which may raise some concerns (which could be discussed in the ADR)
> 
> @mmsqe need to patch the relayer to let it send evm tx instead of cosmos tx.

sorry that I am not so familiar with rust.

thomas-nguy:
```
@mmsqe need to patch the relayer to let it send evm tx instead of cosmos tx.
```

Could it be done without changing the relayer implementation?

something like 

Relayer send cosmos txs -> Cronos middleware catch the cosmos txs and wrap it in an evm txs using module private key 

in a general way, if the concern is to generate the receipt for cosmos txs, I think execution is not needed but we could refactor ethermint (https://github.com/evmos/ethermint/blob/main/x/evm/keeper/msg_server.go#L92) to be able to generate more easily receipts out of an eth txs

yihuang:
> ```
> @mmsqe need to patch the relayer to let it send evm tx instead of cosmos tx.
> ```
> 
> Could it be done without changing the relayer implementation?
> 
> something like
> 
> Relayer send cosmos txs -> Cronos middleware catch the cosmos txs and wrap it in an evm txs using module private key
> 

can you elaborate? by middleware do you mean a rpc service? it can convert the original tx to an evm tx, but can't re-sign the evm tx using the relayer's key, right?

> in a general way, if the concern is to generate the receipt for cosmos txs, I think execution is not needed but we could refactor ethermint (https://github.com/evmos/ethermint/blob/main/x/evm/keeper/msg_server.go#L92) to be able to generate more easily receipts out of an eth txs

there are many things to emulate: tx list in `eth_getBlock`, tx body in `eth_getTransaction`

thomas-nguy:
```
can you elaborate? by middleware do you mean a rpc service? it can convert the original tx to an evm tx, but can't re-sign the evm tx using the relayer's key, right?
```

I mean ibc middleware, or it can even be done in the ante handler if we want to generalise to all cosmos txs

Yes, the evm txs will be signed by an arbitrary key (module key in genesis config) not the relayer key
but I think it is not too important (maybe?) because :
- In case of ibc, what trigger the "change of state" is the relayer, not the original ibc sender and in most of the case, we don't care about the relayer identity (the from field would be irrelevant in most explorer)
- If we need to get the relayer address, we can add an event that will specify it so that it will still be possible to get the oroginal cosmos signer from the explorer

```
there are many things to emulate: tx list in eth_getBlock, tx body in eth_getTransaction, and in the end, we can't emulate the v,r,s signature in tx body.
```

I think we can wrap the cosmos txs in the evm txs like putting the cosmos txs encoded in the call data
 v,r,s signature in tx body will be generated by the module private key
 
 But in some case, we may need to also have the ability to generate "custom" ethereum events which make a bit complex.

yihuang:
> I mean ibc middleware

do you mean broadcast a new tx in the ibc middleware? that is part of the state machine.

thomas-nguy:
`do you mean broadcast a new tx in the ibc middleware? that is part of the state machine.`

One step is to refactor ethermint to make it easier to generate ethereum receipt.

I understand that it is a bit complicated now because its all across many different place. But from my understanding the ethereum receipt information is stored in the cosmos events in many places and reconstructed in the rpc endpoints.

It may be possible to refactor the code to have an utility function to generate receipt. 
(its basically the same as executing ethmsg except that we can ignore the applymsg part)

but maybe there is more complexity (?), I havent look too deep into the code

yihuang:
> `do you mean broadcast a new tx in the ibc middleware? that is part of the state machine.`
> 
> One step is to refactor ethermint to make it easier to generate ethereum receipt.
> 
> I understand that it is a bit complicated now because its all across many different place. But from my understanding the ethereum receipt information is stored in the cosmos events in many places and reconstructed in the rpc endpoints.
> 
> It may be possible to refactor the code to have an utility function to generate receipt. (its basically the same as executing ethmsg except that we can ignore the applymsg part)
> 
> but maybe there is more complexity (?), I havent look too deep into the code

even if we emulated the events, there are still some difficulties, for example `getBlock` filter evm txs by msg type and result code, now it need to look into the event list of every tx, and in the end we still can't emulate the signature of the evm tx.

thomas-nguy:
another possible flow is 

cosmos tx -> middleware convert to eth tx signed with module private key -> execute the eth txs as a specific precompile (we may be able to bypass signature verification check and set direction From to original cosmos sender)

at the end we can get the correct tx result code and gas used. 

but not sure if its possible to change to bypass the signature verification without any risks?

yihuang:
> another possible flow is
> 
> cosmos tx -> middleware convert to eth tx signed with module private key -> execute the eth tx as a specific precompile (we may be able to bypass signature verification check and set direction From to original cosmos sender)
> 
> at the end we can get the correct tx result code and gas used in the receipt.
> 
> but not sure if its possible to change to bypass the signature verification without any risks?

I guess ibc incentivization feature rely on correct signer of the tx.

mmsqe:
@calvinaco it sign normal evm tx, see if we need support dynamic fee tx

x/cronos/keeper/precompiles/relayer.go
```diff
@@ -0,0 +1,162 @@
...
+	BaseContract
+
+	cdc       codec.Codec
+	ibcKeeper *ibckeeper.Keeper
+}
+
+func NewRelayerContract(ibcKeeper *ibckeeper.Keeper, cdc codec.Codec) precompiles.StatefulPrecompiledContract {
+	return &RelayerContract{
+		BaseContract: NewBaseContract(
+			common.BytesToAddress([]byte{101}),
```
yihuang, line 32: define a global constant variable for `101`?

x/cronos/keeper/precompiles/relayer.go
```diff
@@ -0,0 +1,162 @@
...
+		),
+		ibcKeeper: ibcKeeper,
+		cdc:       cdc,
+	}
+}
+
+func (bc *RelayerContract) Address() common.Address {
+	return common.BytesToAddress([]byte{101})
```
yihuang, line 40: https://github.com/crypto-org-chain/cronos/pull/1014/files#r1320933333

x/cronos/keeper/precompiles/relayer.go
```diff
@@ -0,0 +1,162 @@
...
+}
+
+// RequiredGas calculates the contract gas use
+func (bc *RelayerContract) RequiredGas(input []byte) uint64 {
+	// TODO estimate required gas
+	return 0
```
yihuang, line 46: use a constant number for now?

nix/sources.json
```diff
@@ -84,6 +84,18 @@
         "url": "https://github.com/informalsystems/ibc-rs/archive/daad02843a091bbdb3dd608e5f4ce790895c8845.tar.gz",
         "url_template": "https://github.com/<owner>/<repo>/archive/<rev>.tar.gz"
     },
+    "relayer": {
+        "branch": "master",
+        "description": "An IBC relayer for ibc-go",
+        "homepage": "https://github.com/mmsqe/relayer",
```
yihuang, line 90: replace with the fork after merged: https://github.com/crypto-org-chain/relayer/

x/cronos/keeper/precompiles/relayer.go
```diff
@@ -0,0 +1,157 @@
...
+	prefixChannelOpenConfirm
+	prefixChannelCloseInit
+	prefixChannelCloseConfirm
+	prefixRecvPacket
+	prefixAcknowledgement
+	prefixTimeout
+	prefixTimeoutOnClose
+)
+
+// exec is a generic funciton that executes the given action in statedb, and marshal/unmarshal the input/output
```
mmsqe, line 80:
```suggestion
// exec is a generic function that executes the given action in statedb, and marshal/unmarshal the input/output
```

x/cronos/keeper/precompiles/relayer.go
```diff
@@ -0,0 +1,157 @@
...
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	clienttypes "github.com/cosmos/ibc-go/v7/modules/core/02-client/types"
+	conntypes "github.com/cosmos/ibc-go/v7/modules/core/03-connection/types"
+	chantypes "github.com/cosmos/ibc-go/v7/modules/core/04-channel/types"
+	ibckeeper "github.com/cosmos/ibc-go/v7/modules/core/keeper"
+)
+
+// TODO adjust the gas cost
+const RelayerContractRequiredGas = 1000000
```
yihuang, line 22:
```suggestion
const RelayerContractRequiredGas = 10000
```

yihuang (approved):

PR #1163, opened 2023-09-12
Problem: no stateful precompiled contract for ica

mmsqe:

# PR Checklist:


x/cronos/keeper/precompiles/ica.go
```diff
@@ -0,0 +1,199 @@
...
+		args, err := SubmitMsgsMethod.Inputs.Unpack(contract.Input[4:])
+		if err != nil {
+			return nil, errors.New("fail to unpack input arguments")
+		}
+		connectionID := args[0].(string)
+		account := args[1].(common.Address)
+		data := args[2].(string)
+		timeout := args[3].(*big.Int)
+		txSender := evm.Origin
+		if !isSameAddress(account, contract.CallerAddress) && !isSameAddress(account, txSender) {
```
thomas-nguy, line 170: Is `isSameAddress(account, contract.CallerAddress)` enough?

If we enforce owner == tx.origin then we can never call ica from contract address

x/cronos/keeper/precompiles/ica.go
```diff
@@ -0,0 +1,199 @@
...
+	}
+}
+
+func (ic *IcaContract) Address() common.Address {
+	return IcaContractAddress
+}
+
+// RequiredGas calculates the contract gas use
+func (ic *IcaContract) RequiredGas(input []byte) uint64 {
+	return RelayerContractRequiredGas
```
JayT106, line 106: is this TODO in this PR?

mmsqe: yes for now, can use another const for ica

x/cronos/keeper/precompiles/ica.go
mmsqe: need some [change](https://github.com/cosmos/ibc-go/compare/v7.2.0...mmsqe:ibc-go:v7.2.0-rc2?expand=1) in ibc to handle generic

x/cronos/keeper/precompiles/ica.go
JayT106: maybe add a commit for this line for explaining why uses a const instead of calculating the input byte array.

integration_tests/test_upgrade.py
```diff
@@ -184,6 +184,9 @@ def test_cosmovisor_upgrade(custom_cronos: Cronos, tmp_path_factory):
         }
     }
 
+    rsp = cli.query_params("icaauth")
+    assert rsp["params"]["minTimeoutDuration"] == "3600s", rsp
```
mmsqe, line 188:
```suggestion
    assert rsp["params"]["min_timeout_duration"] == "3600s", rsp
```

x/cronos/keeper/precompiles/ica.go
```diff
@@ -0,0 +1,218 @@
...
+		}
+		connectionID := args[0].(string)
+		account := args[1].(common.Address)
+		data := args[2].(string)
+		timeout := args[3].(*big.Int)
+		txSender := evm.Origin
+		if !isSameAddress(account, contract.CallerAddress) && !isSameAddress(account, txSender) {
+			return nil, errors.New("unauthorized send tx")
+		}
+		var icaMsgData icatypes.InterchainAccountPacketData
```
thomas-nguy, line 181: Maybe better to get the msg and construct icatypes.InterchainAccountPacketData from the msg

integration_tests/test_ica_precompile.py
```diff
@@ -0,0 +1,94 @@
...
+    res = contract.caller.nativeQueryAccount(connid, addr)
+    res = re.sub(r"\n>", "", res.decode("utf-8"))
+    assert ica_address == res, res
+
+    funds_ica(cli_host, ica_address)
+    num_txs = len(cli_host.query_all_txs(ica_address)["txs"])
+    str = generate_ica_packet(cli_controller, ica_address, cli_host.address("signer2"))
+    start = w3.eth.get_block_number()
+    # submit transaction on host chain on behalf of interchain account
+    tx = contract.functions.nativeSubmitMsgs(connid, str).build_transaction(data)
```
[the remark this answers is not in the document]
mmsqe: not sure if we need generate for each type like for MsgSend, MsgDelegate...

x/cronos/keeper/precompiles/ica.go
mmsqe: not sure the enforce you mean, but current condition is allow call from CallerAddress or txSender

thomas-nguy: I see, I was concerned that allowing from `tx.origin` may have some security issues but I think should be fine

integration_tests/test_ica_precompile.py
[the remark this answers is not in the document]
thomas-nguy: I think we don't need to generate for all (its up to the contract developer to generate their own proto msgs) but we can provide some boilerplate code for the most useful one such as MsgDelegate etc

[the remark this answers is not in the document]
calvinaco: I have pre-generated the protobuf for Cosmos SDK and IBC here
https://github.com/calvinaco/cosmos-protobuf-solidity

You can just copy one of the module libraries and include it in your example contract

x/cronos/keeper/precompiles/ica.go
```diff
@@ -0,0 +1,218 @@
...
+		}
+		args, err := RegisterAccountMethod.Inputs.Unpack(contract.Input[4:])
+		if err != nil {
+			return nil, errors.New("fail to unpack input arguments")
+		}
+		connectionID := args[0].(string)
+		account := args[1].(common.Address)
+		version := args[2].(string)
+		txSender := evm.Origin
+		if !isSameAddress(account, contract.CallerAddress) && !isSameAddress(account, txSender) {
```
yihuang, line 136: we shouldn't check for `txSender`, because the tx sender may not want to authorize the contract to operate on his behalf.

integration_tests/test_ica_precompile.py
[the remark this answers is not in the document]
calvinaco: Yes, I maintain a fork of the solidity plugin with the fix.

And on the enum what's the change needed? In the pre-generated one I have some rule to sanitize the output because it keep the slash in some imports. I could add more rule to do the manual change
https://github.com/calvinaco/cosmos-protobuf-solidity/blob/master/scripts/generate-protobuf-solidity.sh#L80

integration_tests/contracts/contracts/TestICA.sol
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    function delegateRegister(string memory connectionID) public returns (bool) {
+        (bool result,) = icaContract.delegatecall(encodeRegister(connectionID));
+        require(result, "call failed");
+        return true;
+    }
+
+    function staticRegister(string memory connectionID) public returns (bool) {
+        (bool result,) = icaContract.staticcall(encodeRegister(connectionID));
```
calvinaco, line 27: I am a bit unsure if this is a staiccall or call. The former is not expected to change the blockchain state. But is Cosmos changes a special case?

integration_tests/contracts/contracts/TestICA.sol
```diff
@@ -0,0 +1,81 @@
...
+
+    function encodeQueryAccount(string memory connectionID, address addr) internal view returns (bytes memory) {
+        return abi.encodeWithSignature(
+            "queryAccount(string,address)",
+            connectionID, addr
+        );
+    }
+
+    function callQueryAccount(string memory connectionID, address addr) public returns (string memory) {
+        (bool result, bytes memory data) = icaContract.call(encodeQueryAccount(connectionID, addr));
```
calvinaco, line 40: Similarly here, this appears to be a read only call but instead it is using `call`.

integration_tests/contracts/contracts/ICA.sol
```diff
@@ -0,0 +1,9 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.4;
+
+interface IICAModule {
+    event SubmitMsgsResult(string seq);
+    function registerAccount(string calldata connectionID, string calldata version) external payable returns (bool);
+    function queryAccount(string calldata connectionID, address addr) external view returns (string memory);
+    function submitMsgs(string calldata connectionID, string calldata data, uint256 timeout) external payable returns (uint64);
```
calvinaco, line 8: Here we are still missing a function to `queryPacketSeqResult(string seq) returns (Result)`

The usage is for the smart contract to read the result of previously sent packet sequence and execute subsequent actions if applicable.

mmsqe: Yea, could add one if need

x/cronos/events/bindings/src/ICA.sol
```diff
@@ -0,0 +1,9 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.4;
+
+interface IICAModule {
+    event SubmitMsgsResult(string seq);
+    function registerAccount(string calldata connectionID, string calldata version) external payable returns (bool);
+    function queryAccount(string calldata connectionID, address addr) external view returns (string memory);
+    function submitMsgs(string calldata connectionID, string calldata data, uint256 timeout) external payable returns (uint64);
```
calvinaco, line 8: ditto

x/cronos/events/events.go
```diff
@@ -22,23 +25,45 @@ var (
...
+
+func RelayerConvertEvent(event sdk.Event) (*ethtypes.Log, error) {
+	if event.Type == sdk.EventTypeMessage {
+		return nil, nil
+	}
+	desc, ok := RelayerEvents[event.Type]
+	if !ok {
+		return nil, nil
+	}
+	return desc.ConvertEvent(event.Attributes, RelayerValueDecoders.WithDefaultDecoder(ReturnStringAsIs))
```
yihuang, line 57: I don't like the idea of default decoder, it'll bloat the logs

x/cronos/keeper/precompiles/ica.go
```diff
@@ -0,0 +1,168 @@
...
+	case "queryAccount":
+		args, err := method.Inputs.Unpack(contract.Input[4:])
+		if err != nil {
+			return nil, errors.New("fail to unpack input arguments")
+		}
+		connectionID := args[0].(string)
+		account := args[1].(common.Address)
+		owner := sdk.AccAddress(account.Bytes()).String()
+		icaAddress := ""
+		execErr = stateDB.ExecuteNativeAction(precompileAddr, nil, func(ctx sdk.Context) error {
```
yihuang, line 110: Queries can be done directly, more efficient than `ExecuteNativeAction`, I can add a primitive in ethermint first.

yihuang: here we are: https://github.com/crypto-org-chain/ethermint/pull/359

yihuang (approved):
LGTM, two minor suggestions:
- use new method `CacheContext`  to execute queries
- maybe remove default decoder, and make converted attributes explicit.

x/cronos/events/events.go
mmsqe: Yea, since need relayer change I tend to clean it after this

integration_tests/contracts/contracts/ICA.sol
mmsqe: Seems we need icaauth change for querySubmitMsgsResult, not sure if we can do in another PR

yihuang: It's not ideal to save the acknowledge in state, it bloat the state, and we might need to clean it up later, we can try callbacks.

calvinaco: How does callback works?

> not sure if we can do in another PR
Yes as this PR has certain size already, if you feel like we want to do a cut-off here it's totally fine.

mmsqe: we could try call back precompile in middleware to avoid need save extra data

yihuang: > How does callback works?

call contract method in icaauth ibc module, `OnAcknowledgement` and `OnTimeout`, the contract address is the port-id.

calvinaco: Is it the same as this ADR?
https://github.com/cosmos/ibc-go/blob/v7.3.0/docs/architecture/adr-008-app-caller-cbs/adr-008-app-caller-cbs.md

yihuang: yeah, didn't know about this adr, [v7.3.0](https://github.com/cosmos/ibc-go/releases/tag/v7.3.0) seems perfect for us.

yihuang: I suggest handle the ack part in a separate issue.

x/cronos/keeper/precompiles/ica.go
yihuang: user/contract should provide `data`, precompile should hardcode `type` as `EXECUTE_TX`.

yihuang: fixed

PR #1174, opened 2023-09-19
Problem: not supporting graceful close (fix #1173)

### LINKED ISSUE #1173: Problem: not supporting graceful close
sometimes, with ctrl-c, program not finishes gracefully

leejw51crypto:
wait a little before shutdown

### DIFF
--- a/cmd/cronosd/main.go
+++ b/cmd/cronosd/main.go
@@ -1,7 +1,11 @@
 package main
 
 import (
+	"fmt"
 	"os"
+	"os/signal"
+	"syscall"
+	"time"
 
 	svrcmd "github.com/cosmos/cosmos-sdk/server/cmd"
 	"github.com/crypto-org-chain/cronos/v2/app"
@@ -9,6 +13,23 @@ import (
 )
 
 func main() {
+	// Create a channel to listen for incoming signals
+	sigChan := make(chan os.Signal, 1)
+
+	// Notify the channel for SIGINT and SIGTERM signals
+	signal.Notify(sigChan, syscall.SIGINT, syscall.SIGTERM)
+
+	// Start a goroutine to handle the signal
+	go func() {
+		sig := <-sigChan
+		fmt.Printf("Received signal: %s\n", sig)
+		// Perform cleanup here
+		fmt.Println("Performing cleanup...")
+		time.Sleep(1 * time.Second) // wait for other handlers to finish
+		fmt.Println("Cleanup done!")
+		os.Exit(1)
+	}()
+
 	rootCmd, _ := cmd.NewRootCmd()
 	if err := svrcmd.Execute(rootCmd, cmd.EnvPrefix, app.DefaultNodeHome); err != nil {
 		os.Exit(1)


mmsqe (commented):
just not sure if we could port [this](https://github.com/cosmos/cosmos-sdk/pull/17094/files) one

leejw51crypto:
this is related to versiondb and iavltree  version(height).

cmd/cronosd/main.go
```diff
@@ -1,14 +1,37 @@
...
+	"syscall"
+	"time"
 
 	svrcmd "github.com/cosmos/cosmos-sdk/server/cmd"
 	"github.com/crypto-org-chain/cronos/v2/app"
 	"github.com/crypto-org-chain/cronos/v2/cmd/cronosd/cmd"
 )
 
+import "C"
+
```
mmsqe, line 16:
```suggestion
```

mmsqe (approved):
would be good if we could config 2s

cmd/cronosd/main.go
```diff
@@ -1,14 +1,37 @@
...
 func main() {
+	// Create a channel to listen for incoming signals
+	sigChan := make(chan os.Signal, 1)
+
+	// Notify the channel for SIGINT and SIGTERM signals
+	signal.Notify(sigChan, syscall.SIGINT, syscall.SIGTERM)
+
+	// Start a goroutine to handle the signal
+	go func() {
+		sig := <-sigChan
```
yihuang, line 26: does it work, if main thread quit, it won't wait for the goroutine

leejw51crypto: yes, I'll check further information

yihuang: In ethermint's `server/start.go`, there's some effort to make the quit graceful, but there are some known issues.

yihuang: if you want to try, you can port these changes: https://github.com/cosmos/cosmos-sdk/pull/16203

leejw51crypto: there is 
```
ListenForQuitSignals(g, true, cancelFn, svrCtx.Logger)
```
which also syscall.SIGINT, syscall.SIGTERM, 
so this code, buy a time ListenForQuitSignals settles

yihuang: > so this code, buy a time ListenForQuitSignals settles

why, if main thread quit, the goroutine quit unconditionally, if main thread stays, we don't need the goroutine, right?

yihuang: What the new code does is like this right?
- if main thread can't finish graceful shutdown in 2 seconds, it forcefully shutdown with `os.Exit`
- if main thread shutdown preemptively, it can't do anything.

leejw51crypto:
inspecting ci error

yihuang (changes requested):
Don't understand the point of this change, see https://github.com/crypto-org-chain/cronos/pull/1174#discussion_r1329748898

cmd/cronosd/main.go
leejw51crypto: with latest code, ctrl-c works.
I'll close this pr

leejw51crypto:
with latest branch, ctrl-c works as expected
so close this pr

PR #1178, opened 2023-09-21
Problem: precompiles required-gas need adjustments

thomas-nguy:

Using some base line value for estimation, may need adjustment in the future

Open for suggestions. Keeping this open as it is important to have this fixed before an official release.

# PR Checklist:

### DIFF
--- a/integration_tests/configs/ibc.jsonnet
+++ b/integration_tests/configs/ibc.jsonnet
@@ -5,7 +5,6 @@ config {
     'account-prefix': 'crc',
     'coin-type': 60,
     'precompiled-contract-address': '0x0000000000000000000000000000000000000065',
-    'gas-adjustment': 3,
     key_name: 'signer1',
     accounts: super.accounts[:std.length(super.accounts) - 1] + [super.accounts[std.length(super.accounts) - 1] {
       coins: super.coins + ',100000000000ibcfee',
@@ -140,8 +139,8 @@ config {
     chains: [
       {
         id: 'cronos_777-1',
-        max_gas: 800000,
-        gas_multiplier: 2,
+        max_gas: 500000,
+        gas_multiplier: 4,
         address_type: {
           derivation: 'ethermint',
           proto_type: {
--- a/integration_tests/configs/ibc.jsonnet
+++ b/integration_tests/configs/ibc.jsonnet
@@ -5,6 +5,7 @@ config {
     'account-prefix': 'crc',
     'coin-type': 60,
     'precompiled-contract-address': '0x0000000000000000000000000000000000000065',
+    'gas-adjustment': 3,
     key_name: 'signer1',
     accounts: super.accounts[:std.length(super.accounts) - 1] + [super.accounts[std.length(super.accounts) - 1] {
       coins: super.coins + ',100000000000ibcfee',
@@ -139,7 +140,7 @@ config {
     chains: [
       {
         id: 'cronos_777-1',
-        max_gas: 500000,
+        max_gas: 800000,
         gas_multiplier: 2,
         address_type: {
           derivation: 'ethermint',
--- a/integration_tests/test_ica_precompile.py
+++ b/integration_tests/test_ica_precompile.py
@@ -131,7 +131,7 @@ def test_call(ibc):
     name = "signer2"
     addr = ADDRS[name]
     contract = w3.eth.contract(address=CONTRACT, abi=contract_info)
-    data = {"from": ADDRS[name], "gas": 500000}
+    data = {"from": ADDRS[name], "gas": 400000}
     ica_address = register_acc(
         cli_controller,
         w3,
@@ -173,7 +173,7 @@ def test_sc_call(ibc):
     name = "signer2"
     signer = ADDRS[name]
     keys = KEYS[name]
-    data = {"from": signer, "gas": 500000}
+    data = {"from": signer, "gas": 400000}
     ica_address = register_acc(
         cli_controller,
         w3,
@@ -189,7 +189,7 @@ def test_sc_call(ibc):
 
     # register from another user should fail
     name = "signer1"
-    data = {"from": ADDRS[name], "gas": 500000}
+    data = {"from": ADDRS[name], "gas": 400000}
     version = ""
     tx = tcontract.functions.callRegister(connid, version).build_transaction(data)
     res = send_transaction(w3, tx, KEYS[name])
--- a/integration_tests/test_ica_precompile.py
+++ b/integration_tests/test_ica_precompile.py
@@ -131,7 +131,7 @@ def test_call(ibc):
     name = "signer2"
     addr = ADDRS[name]
     contract = w3.eth.contract(address=CONTRACT, abi=contract_info)
-    data = {"from": ADDRS[name], "gas": 200000}
+    data = {"from": ADDRS[name], "gas": 500000}
     ica_address = register_acc(
         cli_controller,
         w3,
@@ -173,7 +173,7 @@ def test_sc_call(ibc):
     name = "signer2"
     signer = ADDRS[name]
     keys = KEYS[name]
-    data = {"from": signer, "gas": 200000}
+    data = {"from": signer, "gas": 500000}
     ica_address = register_acc(
         cli_controller,
         w3,
@@ -189,7 +189,7 @@ def test_sc_call(ibc):
 
     # register from another user should fail
     name = "signer1"
-    data = {"from": ADDRS[name], "gas": 200000}
+    data = {"from": ADDRS[name], "gas": 500000}
     version = ""
     tx = tcontract.functions.callRegister(connid, version).build_transaction(data)
     res = send_transaction(w3, tx, KEYS[name])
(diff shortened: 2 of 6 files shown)


x/cronos/keeper/precompiles/relayer.go
```diff
@@ -11,8 +11,7 @@ import (
 	ibckeeper "github.com/cosmos/ibc-go/v7/modules/core/keeper"
 )
 
-// TODO adjust the gas cost
-const RelayerContractRequiredGas = 10000
+const RelayerContractRequiredGas = 600000
```
thomas-nguy, line 14: https://cronos.yummy.capital/accounts/crc1tszdy28w2gxldpp0ww6ry58vqhy6rhmx7v6yf6

x/cronos/keeper/precompiles/bank.go
```diff
@@ -18,7 +18,7 @@ import (
 
 const (
 	EVMDenomPrefix          = "evm/"
-	BankContractRequiredGas = 10000
+	BankContractRequiredGas = 21000
```
thomas-nguy, line 21: Estimated to the gas needed for simple transfer (may be increased in the future)

x/cronos/keeper/precompiles/relayer.go
```diff
@@ -11,8 +11,7 @@ import (
 	ibckeeper "github.com/cosmos/ibc-go/v7/modules/core/keeper"
 )
 
-// TODO adjust the gas cost
-const RelayerContractRequiredGas = 10000
+const RelayerContractRequiredGas = 300000
```
thomas-nguy, line 14: https://cronos.yummy.capital/accounts/crc1tszdy28w2gxldpp0ww6ry58vqhy6rhmx7v6yf6

On average, two ibc msgs would used ~600,000 gas. (Update client is the most gas consuming operation with ard 400,000 gas)

x/cronos/keeper/precompiles/bank.go
yihuang: we should apply a consistent mythology to decides the precompile gases, for example do a simulated run of a typical call, collect the cosmos-sdk reported gas, multiply with a coefficient.

thomas-nguy: Correct, If we want to use the cosmos cost as a ground basis, we should probably split all functions within the module precompiles to individual function precompiles (each function within the module could consumes a very different amount of gas) and for ibc we could also top up the cost of relaying (average costs per msgs)

But not sure we would have time to implement it in a short term, if we want to use a static cost then it would be better to bump the required gas to avoid DOS or attack on the relayer

yihuang: the relayer message, it make sense to estimate a gas table for each message type based on production numbers.

calvinaco: > But not sure we would have time to implement it in a short term, if we want to use a static cost then it would be better to bump the required gas to avoid DOS or attack on the relayer

Agree, @yihuang what's your estimation to estimate the gas consumption based on Cosmos cost with coefficient.

If this is considered to be too complex at the moment. I agree we go for a static cost as a version 1, but this cost should be set higher enough to cover all cases but not average to avoid any DOS vector.

x/cronos/keeper/precompiles/relayer.go
calvinaco: Following the above discussions, If we are to hard-coded a static value, I don't think we should take an average and apply to all message types. We should use a value that is always high enough to cover the cost, even if it means we may over-charge the user, as a temporary solution until we change to the Cosmos gas meter.

thomas-nguy: Do you mean the max of all msg?

I think there is a huge difference between `update_client` and `rev_packet`. Could be better to separate in two differents pre compiles

yihuang: We can have a map, and check the prefix of user input

calvinaco: > Do you mean the max of all msg?

Not a single max for all msg, sorry for the confusion. I am expecting each msg has its own max.

JayT106:
Not sure how the caller handles the return `0` from `RequiredGas`. Assume the caller called the contract function with no RequiredGas implementation. Should we return the max of uint64 instead of 0 for the error and the default cases?

x/cronos/keeper/precompiles/bank.go
```diff
@@ -53,8 +56,21 @@ func (bc *BankContract) Address() common.Address {
 
 // RequiredGas calculates the contract gas use
 func (bc *BankContract) RequiredGas(input []byte) uint64 {
-	// TODO estimate required gas
-	return BankContractRequiredGas
+	methodID := input[:4]
```
yihuang, line 59: nit: I think precompute the map from methodID to gas would be more efficient

thomas-nguy:
```
Not sure how the caller handles the return 0 from RequiredGas. Assume the caller called the contract function with no RequiredGas implementation. Should we return the max of uint64 instead of 0 for the error and the default cases?
```

During the execution, the handler would revert because of wrong method name. But it may also consume the gas. Lets say we set to max uint64. This may be a bit too punitive?

On the other hand, this could also be detectable it during the gas estimation phase if not skipped

yihuang (approved):

x/cronos/keeper/precompiles/bank.go
thomas-nguy: done

x/cronos/keeper/precompiles/bank.go
```diff
@@ -48,13 +66,18 @@ func NewBankContract(bankKeeper types.BankKeeper, cdc codec.Codec) vm.Precompile
...
 
 // RequiredGas calculates the contract gas use
 func (bc *BankContract) RequiredGas(input []byte) uint64 {
-	// TODO estimate required gas
-	return BankContractRequiredGas
+	var methodID [4]byte
+	copy(methodID[:], input[:4])
+	requiredGas, ok := bankGasRequiredByMethod[methodID]
+	if ok {
+		return requiredGas
```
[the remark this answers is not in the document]
calvinaco: I agree on adding this. @thomas-nguy wdyt? After this is resolved the rest LGTM.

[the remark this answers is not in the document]
thomas-nguy: added as a basecost

thomas-nguy:
```
 perhaps can charge gas like half of the minimum gas cost of method in the contract?
 ```
 
 Half the minimum of which method? In case method does not exists, there will be no execution on cosmos side. I think charging the base cost should be enough

File: x/cronos/keeper/precompiles/bank.go [hunk 6]
```diff
@@ -18,43 +20,69 @@ import (
...
+	bankABI                 abi.ABI
+	bankContractAddress     = common.BytesToAddress([]byte{100})
+	bankGasRequiredByMethod = map[[4]byte]uint64{}
 )
 
 func init() {
 	if err := bankABI.UnmarshalJSON([]byte(bank.BankModuleMetaData.ABI)); err != nil {
 		panic(err)
 	}
+	for methodName := range bankABI.Methods {
```
mmsqe, line 40: could make gosec happy here

mmsqe (approved):

x/cronos/keeper/precompiles/bank.go
thomas-nguy: the gosec alerts is about the randomness to run over a map right?

not sure how to make it happy

mmsqe: just range warning, btw I see failed to create client error after fee adjust

thomas-nguy: I am suspecting it is a gas issue. 

The go-relayer uses cosmos estimation multiply by an adjustment coef
https://github.com/crypto-org-chain/relayer/blob/main/relayer/chains/cosmos/tx.go#L308C1-L308C1

I have tried using a high adjustement coefficient but no luck


Strangly the test pass locally....

mmsqe: can you fix relayer [config](https://github.com/crypto-org-chain/cronos/blob/78e6423e88295be260751e68313306d8dc563ea2/integration_tests/configs/ibc.jsonnet#L136-L137)

calvinaco (approved):

integration_tests/configs/ibc.jsonnet
```diff
@@ -140,7 +140,7 @@ config {
       {
         id: 'cronos_777-1',
         max_gas: 500000,
-        gas_multiplier: 2,
+        gas_multiplier: 10,
```
mmsqe, line 143: seems a bit too large
```suggestion
        gas_multiplier: 4,
```

integration_tests/test_ica_precompile.py
```diff
@@ -131,7 +131,7 @@ def test_call(ibc):
     name = "signer2"
     addr = ADDRS[name]
     contract = w3.eth.contract(address=CONTRACT, abi=contract_info)
-    data = {"from": ADDRS[name], "gas": 200000}
+    data = {"from": ADDRS[name], "gas": 400000}
```
mmsqe, line 134: Seems we don't need gas change in ica test?

PR #1223, opened 2023-10-25
Problem: no required gas log in precompiles

mmsqe:

# PR Checklist:

### Summary by CodeRabbit

- New Feature: Enhanced logging functionality for the Cronos and ICA contracts, providing more transparency and traceability.
- Refactor: Updated the `baseContract` structure and its associated methods for better configurability and functionality, including a new method to calculate contract gas usage.
- Refactor: Removed the `RequiredGas` function from the `BankContract`, `IcaContract`, and `RelayerContract`, simplifying the gas calculation process.
- New Feature: Introduced method mapping for the bank and ICA contracts, improving the readability and management of contract methods.

These changes aim to improve the overall user experience by enhancing the transparency, traceability, and management of contracts.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
 - [#1215](https://github.com/crypto-org-chain/cronos/pull/1215) Update ethermint to fix of concurrent write in fee history.
 - [#1217](https://github.com/crypto-org-chain/cronos/pull/1217) Use the default chain-id behavour in sdk.
 
+### Improvements
+
+- [#](https://github.com/crypto-org-chain/cronos/pull/) Add required gas log in precompiles.
+
 *October 17, 2023*
 
 ## v1.1.0-rc1
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
 
 ### Improvements
 
-- [#](https://github.com/crypto-org-chain/cronos/pull/) Add required gas log in precompiles.
+- [#1223](https://github.com/crypto-org-chain/cronos/pull/1223) Add required gas log in precompiles.
 
 *October 17, 2023*
 
--- a/x/cronos/keeper/precompiles/base_contract.go
+++ b/x/cronos/keeper/precompiles/base_contract.go
@@ -1,6 +1,8 @@
 package precompiles
 
 import (
+	"github.com/cometbft/cometbft/libs/log"
+	storetypes "github.com/cosmos/cosmos-sdk/store/types"
 	"github.com/ethereum/go-ethereum/common"
 )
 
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,10 @@
 - [#1215](https://github.com/crypto-org-chain/cronos/pull/1215) Update ethermint to fix of concurrent write in fee history.
 - [#1217](https://github.com/crypto-org-chain/cronos/pull/1217) Use the default chain-id behavour in sdk.
 
+### Improvements
+
+- [#](https://github.com/crypto-org-chain/cronos/pull/) Add required gas log in precompiles.
```
mmsqe, line 12:
```suggestion
- [#1223](https://github.com/crypto-org-chain/cronos/pull/1223) Add required gas log in precompiles.
```

x/cronos/keeper/precompiles/base_contract.go
```diff
@@ -10,18 +12,56 @@ type Registrable interface {
...
 }
+
+// RequiredGas calculates the contract gas use
+func (c *baseContract) RequiredGas(input []byte) (gas uint64) {
+	var methodID [4]byte
+	copy(methodID[:], input[:4])
+	inputLen := len(input)
+	defer func() {
+		method := c.nameByMethod[methodID]
+		c.logger.Info("required", "gas", gas, "method", method, "len", inputLen)
```
thomas-nguy, line 56: I think what we need to log is the actual gas consumed during the method execution to compare with the value returned by the method.

The goal is to be able to collect data more easily to tune the "hard-coded" value in the map later on.


Another idea would be to self adjust the gas during the execution (instead of using the hard coded map) but this could lead to some performance degradation

mmsqe: you mean call estimate gas from [relayer](https://github.com/crypto-org-chain/relayer/pull/6/files)?

thomas-nguy: I was more thinking about getting the gas consumed before and after the execution there
https://github.com/crypto-org-chain/cronos/blob/main/x/cronos/keeper/precompiles/bank.go#L132

yihuang, line 56: what's this for, temporary debugging?

mmsqe: I thought we want to record required gas for each method

yihuang: for debugging or what reason?

thomas-nguy: yes debugging, we could collect the data on some nodes to tune the requiredgas map later on

but the data has to be the gas consumed at cosmos level during execution, not the value returned from the map

yihuang: I see, so this could be a draft PR for building a local node, no need to merge, right?

thomas-nguy: I think so, could be just distributed to internal node

However this PR introduces a refactoring which could be nice to merge, maybe change the title of the PR , remove the log and just merge the refactoring?

mmsqe: I tried compare gas used with cosmos and evm tx (current gas map vs estimate via relayer)

  | cosmos | evm_hardcode | evm_estimate
-- | -- | -- | --
MsgCreateClient | 121982 | 240794 | 123774
MsgUpdateClient+MsgConnectionOpenTry | 187503 | 841863 | 250059
MsgUpdateClient+MsgConnectionOpenConfirm | 145679 | 756978 | 177774
MsgUpdateClient+MsgChannelOpenTry | 205426 | 876532 | 240877
MsgUpdateClient+MsgChannelOpenConfirm | 154580 | 775420 | 188570
MsgUpdateClient+MsgRecvPacket | 280213 | 1163360 | 315581

thomas-nguy: Thanks, do you know how cost a single MsgUpdateClient?

Based on that maybe we can use your estimation for better tuning

mmsqe: since rly do auto update client, I get single MsgUpdateClient from hermes

  | herems
-- | --
MsgCreateClient | 122172
MsgConnectionOpenInit | 94511
MsgUpdateClient | 125791
MsgUpdateClient+MsgConnectionOpenAck | 178550
MsgChannelOpenInit | 143607
MsgUpdateClient+MsgChannelOpenAck | 156139
MsgUpdateClient+MsgRecvPacket | 279927

thomas-nguy: I wonder if we could use those data on mainnet
https://cronos.yummy.capital/txs/E2B835B672BC046028E3FEBEB1770CD8A506DB505380982AD16C39F13E5DD861

Somehow hermes in mainnet is spending more gas than that. But may be also because of hermez config?

mmsqe: we should have bigger client header like more validators, and I'm not sure why we use [TransientGasConfig](https://github.com/crypto-org-chain/cronos/blob/main/app/app.go#L548)

thomas-nguy: For the TransientGasConfig, it inspired from evmos implementation of the `gasRequired` 
https://github.com/evmos/evmos/blob/main/precompiles/common/precompile.go#L34

They also use the `TransientGasConfig`, I think that make sense since the goal for us is just to add an "extra" fee based on the length of the input, to avoid DDOS attack for very large input

But in practice, the additional fee does not make much sense since input length is not related to the cost of operation in memory

mmsqe: I think evmos use KvGasConfig instead and input length does related to [TxSizeCost](https://github.com/cosmos/cosmos-sdk/blob/v0.47.5/x/auth/ante/basic.go#L96) which is the major cost in cosmos tx

mmsqe: cosmos tx gas can be trace from relayer without precompile config, the gas meter consumed is same as the log from [RequiredGas](https://github.com/crypto-org-chain/cronos/pull/1223/files#diff-b93d562e3e27735b0f3311c3b6cf9b278f98f5a1fe03b09cfe232ef752365497R100)

thomas-nguy: mmm in the original context, the base cost is just an extra cost to avoid DDOS attack, since the input it never written in store, it would make sense to use transient store cost for this case.

But I think what you are trying to do is to replicate the cosmos intrinsic cost? In that case KvGasConfig is used

PR #1232, opened 2023-10-31
Problem: require gas in relayer precompile is higher than consumed

mmsqe:

# PR Checklist:

## Summary by CodeRabbit

- **Bug Fixes**
	- Fixed a memory store issue improving overall performance.
	- Adjusted the required gas in the relayer precompile for accurate gas consumption.

- **New Features**
	- Added a new chain 'chainmain-1' with two validators in the integration tests.
	- Introduced a method for transaction search based on given criteria.
	- Added a function for token transfer between two chains using the command-line tool.
	- Implemented a new test function to test Inter-Blockchain Communication (IBC) functionality.

- **Refactor**
	- Updated the `NewRelayerContract` function to require a logger as an argument.
	- Modified the `RelayerContract` struct and its associated functions for better gas usage calculation.
	- Added new functions to the `IbcKeeper` interface.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,11 @@
 - [#1230](https://github.com/crypto-org-chain/cronos/pull/1230) Fix mem store in versiondb multistore.
 - [#1233](https://github.com/crypto-org-chain/cronos/pull/1233) Re-emit logs in callback contract.
 
+### State Machine Breaking
+
+- [#1232](https://github.com/crypto-org-chain/cronos/pull/1232) Adjust require gas in relayer precompile to be closed with actual consumed.
+
+
 *October 17, 2023*
 
 ## v1.1.0-rc1
--- a/x/cronos/keeper/precompiles/relayer.go
+++ b/x/cronos/keeper/precompiles/relayer.go
@@ -34,7 +34,7 @@ func init() {
 	relayerGasRequiredByMethod[prefixChannelOpenTry] = 70562
 	relayerGasRequiredByMethod[prefixChannelOpenAck] = 22127
 	relayerGasRequiredByMethod[prefixChannelOpenConfirm] = 21190
-	relayerGasRequiredByMethod[prefixRecvPacket] = 144925
+	relayerGasRequiredByMethod[prefixRecvPacket] = 144025
 	relayerGasRequiredByMethod[prefixAcknowledgement] = 61781
 	relayerGasRequiredByMethod[prefixTimeout] = 104283
 	relayerGasRequiredByMethod[prefixTimeoutOnClose] = 100000
--- a/x/cronos/keeper/precompiles/relayer.go
+++ b/x/cronos/keeper/precompiles/relayer.go
@@ -64,6 +64,7 @@ func (bc *RelayerContract) Address() common.Address {
 }
 
 // RequiredGas calculates the contract gas use
+// `len(input) * DefaultTxSizeCostPerByte + requiredGasTable[methodPrefix] - intrinsicGas`
 func (bc *RelayerContract) RequiredGas(input []byte) (gas uint64) {
 	if len(input) < prefixSize4Bytes {
 		return 0
--- a/x/cronos/keeper/precompiles/relayer.go
+++ b/x/cronos/keeper/precompiles/relayer.go
@@ -64,7 +64,7 @@ func (bc *RelayerContract) Address() common.Address {
 }
 
 // RequiredGas calculates the contract gas use
-// `len(input) * DefaultTxSizeCostPerByte + requiredGasTable[methodPrefix] - intrinsicGas`
+// `max(0, len(input) * DefaultTxSizeCostPerByte + requiredGasTable[methodPrefix] - intrinsicGas)`
 func (bc *RelayerContract) RequiredGas(input []byte) (gas uint64) {
 	if len(input) < prefixSize4Bytes {
 		return 0
--- a/x/cronos/keeper/precompiles/relayer.go
+++ b/x/cronos/keeper/precompiles/relayer.go
@@ -80,8 +80,6 @@ func (bc *RelayerContract) RequiredGas(input []byte) (gas uint64) {
 	}
 	// base cost to prevent large input size
 	baseCost := uint64(len(input)) * authtypes.DefaultTxSizeCostPerByte
-	var methodID [4]byte
-	copy(methodID[:], input[:4])
 	defer func() {
 		bc.logger.Debug("required", "gas", gas, "method", prefix, "len", len(input), "intrinsic", intrinsicGas)
 	}()
--- a/x/cronos/keeper/precompiles/relayer.go
+++ b/x/cronos/keeper/precompiles/relayer.go
@@ -34,6 +34,8 @@ func init() {
 	relayerGasRequiredByMethod[prefixChannelOpenTry] = 70562
 	relayerGasRequiredByMethod[prefixChannelOpenAck] = 22127
 	relayerGasRequiredByMethod[prefixChannelOpenConfirm] = 21190
+	relayerGasRequiredByMethod[prefixChannelCloseInit] = 100000
+	relayerGasRequiredByMethod[prefixChannelCloseConfirm] = 31199
 	relayerGasRequiredByMethod[prefixRecvPacket] = 144025
 	relayerGasRequiredByMethod[prefixAcknowledgement] = 61781
 	relayerGasRequiredByMethod[prefixTimeout] = 104283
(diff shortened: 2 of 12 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,11 @@
 - [#1216](https://github.com/crypto-org-chain/cronos/pull/1216) Update ethermint to fix of avoid redundant parse chainID from gensis when start server.
 - [#1230](https://github.com/crypto-org-chain/cronos/pull/1230) Fix mem store in versiondb multistore.
 
+### Improvements
+
+- [#](https://github.com/crypto-org-chain/cronos/pull/) Adjust require gas in relayer precompile to be closed with actual consumed.
```
mmsqe, line 14:
```suggestion
- [#1232](https://github.com/crypto-org-chain/cronos/pull/1232) Adjust require gas in relayer precompile to be closed with actual consumed.
```

x/cronos/keeper/precompiles/relayer.go
```diff
@@ -60,18 +62,31 @@ func (bc *RelayerContract) Address() common.Address {
...
 
 // RequiredGas calculates the contract gas use
-func (bc *RelayerContract) RequiredGas(input []byte) uint64 {
-	// base cost to prevent large input size
-	baseCost := uint64(len(input)) * bc.kvGasConfig.WriteCostPerByte
+func (bc *RelayerContract) RequiredGas(input []byte) (gas uint64) {
 	if len(input) < prefixSize4Bytes {
 		return 0
 	}
+	intrinsicGas, err := core.IntrinsicGas(input, nil, false, true, true)
```
thomas-nguy, line 69: I'm not sure what is the logic there to compute the intrinsic gas but the gas cost should be related to the cosmos cost (ie storage ops), not the ethereum one?

x/cronos/keeper/precompiles/relayer.go
```diff
@@ -60,18 +62,31 @@ func (bc *RelayerContract) Address() common.Address {
...
+	copy(methodID[:], input[:4])
+	defer func() {
+		bc.logger.Debug("required", "gas", gas, "method", prefix, "len", len(input), "intrinsic", intrinsicGas)
+	}()
+	total := requiredGas + baseCost
+	if total < intrinsicGas {
+		return 0
 	}
-	return baseCost
+	return total - intrinsicGas
```
thomas-nguy, line 89: why returning the diff instead of the minimum or something more reasonable?

x/cronos/keeper/precompiles/relayer.go
mmsqe: patch to be closed how [eth](https://github.com/ethereum/go-ethereum/blob/v1.10.26/core/state_transition.go#L118) calculate gas, while [cosmos](https://github.com/cosmos/cosmos-sdk/blob/v0.47.5/x/auth/ante/basic.go#L96) only use len(input) * 10, you could check Details [here](https://github.com/crypto-org-chain/cronos/pull/1209#issue-1944175530)

thomas-nguy: Got it.
I think the question is more if ETH gas calculation is relevant here or if we should only use cosmos computation method.

mmsqe: we will have bigger gap since intrinsic gas is charged when sign evm tnx, while cosmos doesn't have different price for zero and non-zero.

thomas-nguy: I still don't understand the logic of returning `total - intrinsicGas`

Per definition `The intrinsic gas for a transaction is the amount of gas that the transaction uses before any code run`

In that case should we replace the baseCost value by the intrinsic gas value?

mmsqe: intrinsic gas will be charged, it's to minimize the diff between cosmos and eth?

thomas-nguy: I see,

In that case, in that case, does it make sense to keep the intrinsic gas as "extra charge" since evm operations is more resource consuming that cosmos native operation?

mmsqe: we call cosmos msg server in precompile, and won't we keep gas close to cosmos cost? intrinsic has bigger variant based on ratio of zero and non-zero, can you explain your formula?

x/cronos/keeper/precompiles/relayer.go
```diff
@@ -60,18 +64,31 @@ func (bc *RelayerContract) Address() common.Address {
...
+	}
 	prefix := int(binary.LittleEndian.Uint32(input[:prefixSize4Bytes]))
 	requiredGas, ok := relayerGasRequiredByMethod[prefix]
-	if ok {
-		return requiredGas + baseCost
+	if !ok {
+		requiredGas = 0
+	}
+	// base cost to prevent large input size
+	baseCost := uint64(len(input)) * authtypes.DefaultTxSizeCostPerByte
```
yihuang, line 81: would a simple `len(input) * a + b` formula has as good approximation?

mmsqe: you mean ignore intrinsic like old method right? we'll get `(24196 * 10 + 0) + 405232 - 392132 = 255060` diff



90 validtors | cosmos | evm | adjust | current | input | intrinsic | update client | 2nd method
-- | -- | -- | -- | -- | -- | -- | -- | --
MsgUpdateClient+MsgConnectionOpenTry | 392132 | 405232 | 590472 | 1611456 | 24196 | 405232 | 429428 | 834660

yihuang: pick a better `a` ofc, I assume intrinsic gas is more or less linearly correlated with input length.

mmsqe: Actually intrinsic varies with input, let's use TransientGasConfig 3 as before, we'll have negative with smaller nodes `(3684 * 10 + 0) + 77712 - 187154 = -72602`


2 validtors | cosmos | evm | adjust | current | input | intrinsic | update client | 2nd method
-- | -- | -- | -- | -- | -- | -- | -- | --
MsgUpdateClient+MsgConnectionOpenTry | 187154 | 187202 | 232850 | 808302 | 3684 | 77712 | 81396 | 159108

x/cronos/keeper/precompiles/relayer.go
```diff
@@ -60,18 +64,31 @@ func (bc *RelayerContract) Address() common.Address {
 }
 
 // RequiredGas calculates the contract gas use
-func (bc *RelayerContract) RequiredGas(input []byte) uint64 {
-	// base cost to prevent large input size
-	baseCost := uint64(len(input)) * bc.kvGasConfig.WriteCostPerByte
+func (bc *RelayerContract) RequiredGas(input []byte) (gas uint64) {
```
yihuang, line 67:
```suggestion
// `len(input) * DefaultTxSizeCostPerByte + requiredGasTable[methodPrefix] - intrinsicGas`
func (bc *RelayerContract) RequiredGas(input []byte) (gas uint64) {
```

x/cronos/keeper/precompiles/relayer.go
thomas-nguy: Precompile is a cosmos msg executed in an evm environment right?

In that case it should be most costly than a standard cosmos msg which do not need evm initialisation. 

Basically per definition every evm msg needs to pay the intrinsic gas cost. since calling precompile is an evm msg, I don't know why we remove this cost

x/cronos/keeper/precompiles/relayer.go
yihuang: Solve this equation:
```
24196 * a + b + 405232 = 392132
3684 * a + b + 77712 = 187154
```

we get `a ~= -6, b ~= 131451`

x/cronos/keeper/precompiles/relayer.go
thomas-nguy: But I agree that since we already pay the intrinsic cost per default, we should probably remove the baseCost w

thomas-nguy: But I agree that since we already pay the intrinsic cost per default, we should probably remove the baseCost which may be redundant. Any feedback @yihuang ?

x/cronos/keeper/precompiles/relayer.go
mmsqe: yea, require [0](https://github.com/crypto-org-chain/cronos/blob/7b3c0f4a39bde4db132340e12b300326725b0092/x/cronos/keeper/precompiles/relayer.go#L90) in that case

yihuang (approved):

x/cronos/keeper/precompiles/relayer.go
yihuang: LGTM based on the resulting approximation actually.

x/cronos/keeper/precompiles/relayer.go
```diff
@@ -60,18 +64,32 @@ func (bc *RelayerContract) Address() common.Address {
 }
 
 // RequiredGas calculates the contract gas use
-func (bc *RelayerContract) RequiredGas(input []byte) uint64 {
-	// base cost to prevent large input size
-	baseCost := uint64(len(input)) * bc.kvGasConfig.WriteCostPerByte
+// `len(input) * DefaultTxSizeCostPerByte + requiredGasTable[methodPrefix] - intrinsicGas`
```
yihuang, line 67:
```suggestion
// `max(0, len(input) * DefaultTxSizeCostPerByte + requiredGasTable[methodPrefix] - intrinsicGas)`
```

x/cronos/keeper/precompiles/relayer.go
```diff
@@ -60,18 +64,32 @@ func (bc *RelayerContract) Address() common.Address {
...
 	requiredGas, ok := relayerGasRequiredByMethod[prefix]
-	if ok {
-		return requiredGas + baseCost
+	if !ok {
+		requiredGas = 0
+	}
+	// base cost to prevent large input size
+	baseCost := uint64(len(input)) * authtypes.DefaultTxSizeCostPerByte
+	var methodID [4]byte
+	copy(methodID[:], input[:4])
```
yihuang, line 84:
```suggestion
```

not used?

mmsqe: no

PR #1851, opened 2025-08-20
fix: Add default value of CancunTime and PragueTime in chain config

XinyuCRO:
ethermint pr: https://github.com/crypto-org-chain/ethermint/pull/702

## Summary by CodeRabbit

* **Documentation**
  * Updated UNRELEASED notes to document default CancunTime and PragueTime and minor formatting cleanup.

* **Chores**
  * Bumped a core dependency snapshot and updated its lock/hash.
  * Set default Cancun/Prague values and added blob scheduling entries in genesis configuration.

* **Tests**
  * Added helpers to fund test accounts and to remove Cancun/Prague params via governance.
  * Expanded integration tests for pre- and post-Cancun SELFDESTRUCT behavior.

* **Contracts (tests)**
  * Added payable receive handlers to test contracts to support new suicide tests.

### DIFF
--- a/integration_tests/test_basic.py
+++ b/integration_tests/test_basic.py
@@ -25,11 +25,13 @@ from .utils import (
     contract_path,
     deploy_contract,
     derive_new_account,
+    fund_acc,
     get_account_nonce,
     get_expedited_params,
     get_receipts_by_block,
     get_sync_info,
     modify_command_in_supervisor_config,
+    remove_cancun_prague_params,
     replace_transaction,
     send_transaction,
     send_txs,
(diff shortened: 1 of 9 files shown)


integration_tests/test_basic.py
```diff
@@ -684,16 +684,21 @@ def test_suicide(cluster):
...
     tx = destroyer.functions.check_codesize_after_suicide(
         destroyee.address
     ).build_transaction()
     receipt = send_transaction(w3, tx)
     assert receipt.status == 1
 
-    assert len(w3.eth.get_code(destroyee.address)) == 0
+    # per EIP-6780, SELFDESTRUCT will recover all funds to the target
+    # but not delete the account, except when called in the same transaction as creation
+    assert w3.eth.get_code(destroyee.address) == old_code_destroyee
```
calvinaco, line 733: I think it is better for us to have two separate test cases to test pre EIP-6780 and post EIP-6780 behaviour by enabling/disabling Cancun and Prague.

XinyuCRO: added, but to make geth in integration tests act before cancun requires more work within the current test setup, so the pre cancun selfdestruct test only involves cronos

PR #1948, opened 2026-01-05
chore(attestation): clean up and fix db lock

JayT106:

# PR Checklist:

### DIFF
--- a/x/attestation/flags.go
+++ b/x/attestation/flags.go
@@ -1,6 +1,5 @@
 package attestation
 
-
 import (
 	"github.com/spf13/cobra"
 )
@@ -9,7 +8,7 @@ const (
 	// FlagDAIBCVersion is the flag to specify the IBC version for data availability
 	FlagDAIBCVersion = "da-ibc-version"
 
-	// Default IBC version
+	// DefaultDAIBCVersion is the default IBC version for data availability attestation
 	DefaultDAIBCVersion = "v2"
 )
 
--- a/x/attestation/keeper/ibc_module_v1.go
+++ b/x/attestation/keeper/ibc_module_v1.go
@@ -171,7 +171,9 @@ func (im IBCModuleV1) OnChanCloseConfirm(
 	channelID string,
 ) error {
 	// Store the channel ID for future use
-	im.keeper.SetV1ChannelID(ctx, "attestation-layer", channelID)
+	if err := im.keeper.SetV1ChannelID(ctx, "attestation-layer", channelID); err != nil {
+		return errorsmod.Wrapf(err, "failed to set v1 channel ID")
+	}
 
 	return nil
 }
--- a/x/attestation/keeper/local_storage_simple_test.go
+++ b/x/attestation/keeper/local_storage_simple_test.go
@@ -124,6 +124,7 @@ func TestLocalStorageWithRealDB(t *testing.T) {
 }
 
 func testLocalDB(t *testing.T, dbPath string) {
+	t.Helper()
 	// Test that we can create and close a database
 	// This is a basic sanity check
 


app/app.go
```diff
@@ -1121,6 +1118,17 @@ func New(
 			app.Logger().Error("failed to update blocklist", "error", err)
 		}
 
+		// Initialize local finality storage (non-consensus)
+		// Only initialize if this is a real node (has app.toml config) not a temp app
+		// used for CLI setup. Temp apps use simtestutil.NewAppOptionsWithFlagHome which
+		// only sets FlagHome, so we check for a real server option like "api.enable".
+		// The real server loads config from app.toml which sets this option.
+		if appOpts.Get("api.enable") != nil {
```
randy-cro: i think just pass in the db type from the earlier `New` function. the tempapp will pass in an in memdb, then the lock issue will be resolved, no need for any checks. for `setupAttestationFinalityStorage` function, just pass in the db parameter, there is no need to read the flag again?

x/attestation/module.go
```diff
@@ -416,14 +401,14 @@ func (am AppModule) collectBlockAttestations(ctx context.Context, startHeight, e
...
-	am.keeper.Logger(ctx).Info("collected block attestations from collector",
+	am.keeper.Logger(ctx).Debug("collected block attestations from collector",
 		"start_height", startHeight,
 		"end_height", endHeight,
 		"count", len(attestations),
 	)
 
 	for _, attestation := range attestations {
-		am.keeper.Logger(ctx).Info("GONNA SEND block attestation data",
+		am.keeper.Logger(ctx).Debug("GONNA SEND block attestation data",
```
randy-cro: can remove this, this was my debug log. if wanna keep as debug log, can rephrase it instead of `GONNA SEND block attestation data`

JayT106: Removed.

app/app.go
JayT106: Fixed!

app/app.go
```diff
@@ -808,11 +808,13 @@ func New(
 	app.IBCKeeper.SetRouterV2(ibcV2Router)
 
 	// Initialize local finality storage (non-consensus)
-	if err := setupAttestationFinalityStorage(app, homePath, appOpts, logger); err != nil {
+	if err := setupAttestationFinalityStorage(app, homePath, appOpts, logger, db); err != nil {
 		logger.Warn("Failed to initialize attestation local finality storage", "error", err)
 	}
 
 	// Store CometBFT RPC address for later initialization (after ABCI handshake)
+	// NOTE: Attestation finality storage initialization is deferred to loadLatest block
```
randy-cro: these 2 lines of comments, not needed?

app/attestation.go
```diff
@@ -18,13 +19,22 @@ import (
...
+// If db is provided (e.g., MemDB for tempApp), it will be used directly instead of creating disk-based storage.
+// If db is nil, a persistent database will be created at homePath/data/finality.
+func setupAttestationFinalityStorage(app *App, homePath string, appOpts servertypes.AppOptions, logger log.Logger, db dbm.DB) error {
 	// Get cache size from config (default: 10000)
 	cacheSize := cast.ToInt(appOpts.Get("attestation.finality-cache-size"))
 	if cacheSize == 0 {
 		cacheSize = 10000
 	}
 
+	if db == nil || reflect.TypeOf(db).String() == "*db.MemDB" {
```
randy-cro: not sure what is this for,  im thinking we can just use the db directly and remove line 40. just use the db supplied from the argument and do all the checks? im thinking we should just log an error should the db be nil?

JayT106: Because from the chain setup script. the tempApp using MemDB, it will call NewApp function and then call this function. But I think I can just check it before call `setupAttestationFinalityStorage`, and return an error if db is nil and dbtype is MemDB (which it doesn't be supported the attestation feature).

x/attestation/collector/collector.go
```diff
@@ -163,13 +160,13 @@ func (c *BlockDataCollector) collectLoop(ctx context.Context, eventCh <-chan cty
 }
 
 // collectAndStoreBlock collects full block data and stores it locally
-func (c *BlockDataCollector) collectAndStoreBlock(ctx context.Context, eventData cmttypes.EventDataNewBlock) error {
+func (c *BlockDataCollector) collectAndStoreBlock(eventData cmttypes.EventDataNewBlock) error {
```
thomas-nguy, line 163: The block data should already be stored by cometBFT isnt it? Any reason for the duplicate?

JayT106: good find, after the protocol changes, we don't need to collect blockdata and stores, I will remove collector and submit another PR to address it?

JayT106: https://github.com/crypto-org-chain/cronos/pull/1949

app/app.go
```diff
@@ -808,8 +808,12 @@ func New(
 	app.IBCKeeper.SetRouterV2(ibcV2Router)
 
 	// Initialize local finality storage (non-consensus)
-	if err := setupAttestationFinalityStorage(app, homePath, appOpts, logger); err != nil {
-		logger.Warn("Failed to initialize attestation local finality storage", "error", err)
+	_, isMemDB := db.(*dbm.MemDB)
+
+	if !isMemDB {
+		if err := setupAttestationFinalityStorage(app, homePath, appOpts, logger, db); err != nil {
```
randy-cro: 1. i feel the `memDB` check is not needed, unless creating in memory attestation finality storage costs alot.
2. Also seems like the db parameter passed in is not even used? `setupAttestationFinalityStorage` should use the db passed in directly for the database creation instead of reading from `app-db-backend` again?

PR #1982, opened 2026-03-10
fix(nix): enable LZ4 support for Windows builds

songgaoye:

# PR Checklist:
summary
- When RocksDB is built without LZ4 support, opening a DB containing LZ4-compressed blocks fails at runtime with:
  "Not implemented: LZ4 not supported in this build".
- After enabling WITH_LZ4 on MinGW, RocksDB must also link against LZ4 correctly. On MinGW, linking to a DLL typically relies on an import library (liblz4.dll.a).
- The nixpkgs MinGW lz4 package can provide a stub/wrong import library, so RocksDB compiles LZ4 code paths but the link step fails (undefined reference to LZ4_* symbols), or we end up effectively not linking LZ4 even though WITH_LZ4 is enabled.
What this PR does
- Forces RocksDB builds to include LZ4 on Windows/MinGW (so mainnet DBs that use LZ4 can be opened).
- For MinGW specifically, regenerates a correct import library from the real liblz4.dll exports:
  1) locate the actual liblz4.dll
  2) extract exported LZ4_* symbols from `objdump -p` output
  3) write a .def file
  4) run dlltool to produce liblz4.dll.a
  5) point RocksDB (via pkg-config / CMake flags) at the regenerated import library
- Adds a fail-fast guard and prints `objdump -p` output if export extraction yields no LZ4 symbols, to avoid silently generating an empty import lib.

- install Wine, create LZ4 RocksDB fixture, run version + db migrate under Wine

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@
 
 ### Bug fixes
 
-* [#1982] (https://github.com/crypto-org-chain/cronos/pull/1982) fix(nix): enable LZ4 support for Windows builds.
+* [#1982](https://github.com/crypto-org-chain/cronos/pull/1982) fix(nix): enable LZ4 support for Windows builds.
 * [#1946](https://github.com/crypto-org-chain/cronos/pull/1946) Fix: allow preinstall in case account has a balance
 * [#1947](https://github.com/crypto-org-chain/cronos/pull/1947) Fix: estimate gas not accurate
 * [#1929](https://github.com/crypto-org-chain/ethermint/pull/1929) fix: add check on evm transaction tip


songgaoye:
> LTGM. Have you tested the window build can be executed correctly? Or should we add github CI task to check it.

add check

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@
 
 ### Bug fixes
 
+* [#1982] (https://github.com/crypto-org-chain/cronos/pull/1982) fix(nix): enable LZ4 support for Windows builds.
```
XinyuCRO:
```suggestion
* [#1982](https://github.com/crypto-org-chain/cronos/pull/1982) fix(nix): enable LZ4 support for Windows builds.
```

PR #2118, opened 2026-06-23
perf(mempool): run post-Commit recheck async off consensus path

JayT106:
### Problem

`App.Commit()` called `RecheckTxs()` synchronously on every block. That function scans the whole mempool and reruns every affected transaction through `RunTx(ReCheck)`. Running that on the consensus-critical path slowed down the validator's next round.

### Fix

- `Commit()` now just fires a non-blocking trigger. A background worker picks it up and does the recheck off the hot path.
- If several triggers land in the same window, they collapse into one run, but the senders that need rechecking are still merged in — none get dropped.
- `PrepareProposal` waits up to 1.5 seconds for that recheck to finish before picking transactions, so it isn't building a block from a stale pool. If the recheck is still running past that, it returns an empty proposal instead of guessing.
- `Close()` now waits for the worker to finish its current recheck before the store shuts down, so nothing reads from a closed store.
- Fixed a bug where the timeout check read the context's error after canceling it, which always reported "timed out" even when the recheck had actually finished in time. Pulled that logic into its own function (`WaitForRecheckTimedOut`) and added tests for both outcomes.
- Fixed a race where the worker could quit while a trigger was still queued, which used to leave a caller waiting forever. Added a regression test for it.
- Added a test confirming shutdown can be called twice safely.
- Fixed a bug in `runRecheck` where a transaction that failed `RunTx(ReCheck)` was only removed from the encoder cache, not the pool. BaseApp only auto-removes a transaction from the pool when the ante handler fails recheck; if the failure happens later, during message execution, the transaction stayed in the pool and could get reselected into a future proposal. This predates this PR but surfaced during review, so it's fixed here too, with a regression test covering the msg-execution-failure path specifically.


app/mempool/manager.go
```diff
@@ -243,13 +257,53 @@ func (a *Manager) StageRecheckSenders(height int64, txs [][]byte) {
...
+
+// recheckWorker runs RecheckTxs off the Commit path. Single goroutine, so
+// worker-driven rechecks never overlap.
+func (a *Manager) recheckWorker() {
+	defer close(a.workerDone)
+	for {
+		select {
+		case <-a.stop:
+			return
+		case <-a.recheckTrigger:
```
songgaoye: `recheckWorker` should re-check stop after receiving `recheckTrigger`. If stop is closed while a trigger is buffered, Go select may choose the trigger case and start another full RechecksTx during shutdown, making App.Close block unnecessarily store teardown.

app/mempool/manager.go
```diff
@@ -243,13 +257,53 @@ func (a *Manager) StageRecheckSenders(height int64, txs [][]byte) {
 	a.stagingMu.Unlock()
 }
 
+// TriggerRecheck schedules an async recheck off the consensus path. Non-blocking:
+// a queued run is reused (staging merges across blocks, so no senders are lost).
+// Inline when no worker (built via newManager).
+func (a *Manager) TriggerRecheck() {
```
songgaoye, line 254: The next `PrepareProposal` may run before the async worker evicts stale txs. Since the fast proposal path skips proposal-time ante, stale nonce/balance/timeout txs can enter proposals

app/mempool/manager.go
```diff
@@ -243,13 +257,53 @@ func (a *Manager) StageRecheckSenders(height int64, txs [][]byte) {
...
+		return
+	}
+	a.stopOnce.Do(func() { close(a.stop) })
+	<-a.workerDone
+}
+
 // RecheckTxs evicts pool txs invalidated by the last block: timed-out txs (any
 // sender) and txs of block-touched senders that now fail ReCheck. Capped per
 // cycle; overflow carries to the next.
 func (a *Manager) RecheckTxs() {
```
songgaoye, line 287: Could you add some async lifecycle tests such as
`
TriggerRecheck wakes worker
coalesced triggers preserve staged senders
multiple commits while recheck in running
Commit -> immediate PrepareProposal
Close waits for in-flight recheck
`

app/mempool/manager.go
```diff
@@ -243,13 +257,53 @@ func (a *Manager) StageRecheckSenders(height int64, txs [][]byte) {
...
 	if a.mpool == nil {
 		return
 	}
+	a.recheckMu.Lock() // single-writer; see recheckMu
```
songgaoye: Thanks, we discussed offline that the admission/recheck ownership split will be handled in a follow-up. That sounds good to me.
For this PR, I only want to make sure the scope is clear: `recheckMu` serializes `RecheckTxs` and protects recheck-local state like `arrival`, but it does not by itself solve the existing BaseApp `checkState` interleaving issue inside `runRecheck`.

app/mempool/manager.go
```diff
@@ -52,11 +52,19 @@ type Manager struct {
...
 	// ttlNumBlocks evicts txs older than this many blocks by arrival height; 0 = off.
 	ttlNumBlocks int64
+
+	// recheckMu serializes RecheckTxs so the lock-free arrival map has one writer.
+	// Always acquired before mu/stagingMu, never after.
+	recheckMu sync.Mutex
+	// recheckTrigger wakes the async worker (buffered-1, coalescing); nil in tests (sync path).
+	recheckTrigger chan struct{}
+	stop           chan struct{}
+	workerDone     chan struct{}
```
thomas-nguy: If they are used only for recheck, please use a better naming to avoid confusion. Could be move to another object as well

app/mempool/manager.go
thomas-nguy: That is exactly my thinking reading the change. 
How to you prevent stale txs to be included in the next block?
The prepare proposal needs to be aware about the cleaning job and make sure recheck has been executed before starting a new block

app/mempool/manager.go
JayT106: moved to an object dedicated for async recheck

app/mempool/manager.go
JayT106: fixed.

app/mempool/manager.go
JayT106: done

JayT106:
@claude review

app/mempool/manager.go
JayT106: added `WaitForRecheck` with ready channel to gate the prepare proposal.

app/mempool/manager.go
```diff
@@ -52,11 +52,24 @@ type Manager struct {
...
 	// arrival maps each pooled tx to the height RecheckTxs first observed it, for
-	// ttlNumBlocks eviction. Rebuilt from the snapshot each cycle (stale entries
-	// drop out) and touched only by RecheckTxs (serial per Commit), so it needs no lock.
+	// ttlNumBlocks eviction. Rebuilt from the snapshot each cycle; recheckMu keeps it single-writer.
 	arrival map[sdk.Tx]int64
 	// ttlNumBlocks evicts txs older than this many blocks by arrival height; 0 = off.
 	ttlNumBlocks int64
+
+	recheckMu sync.Mutex // serializes RecheckTxs; acquired before mu/stagingMu, never after
+	// zero in the test path (newManager); sync recheck used there.
```
thomas-nguy: what does it mean?

app/mempool/manager.go
```diff
@@ -52,11 +52,24 @@ type Manager struct {
...
+	async recheckAsync
+}
+
+// recheckAsync holds lifecycle state for the async recheck worker.
+type recheckAsync struct {
```
thomas-nguy: The recheck logic is all over the place, why don't you create an object recheck worker and move all the logic there?

The flow would be easier to understand if you split the logic

thomas-nguy:
```suggestion
type recheckWorker struct {
```

app/mempool/manager.go
```diff
@@ -52,11 +52,24 @@ type Manager struct {
...
-	// ttlNumBlocks eviction. Rebuilt from the snapshot each cycle (stale entries
-	// drop out) and touched only by RecheckTxs (serial per Commit), so it needs no lock.
+	// ttlNumBlocks eviction. Rebuilt from the snapshot each cycle; recheckMu keeps it single-writer.
 	arrival map[sdk.Tx]int64
 	// ttlNumBlocks evicts txs older than this many blocks by arrival height; 0 = off.
 	ttlNumBlocks int64
+
+	recheckMu sync.Mutex // serializes RecheckTxs; acquired before mu/stagingMu, never after
+	// zero in the test path (newManager); sync recheck used there.
+	async recheckAsync
```
thomas-nguy:
```suggestion
	worker recheckWorker
```

app/mempool/manager.go
```diff
@@ -243,13 +263,78 @@ func (a *Manager) StageRecheckSenders(height int64, txs [][]byte) {
...
+	case a.async.trigger <- ready:
+		a.async.ready = ready // track the latest gate for WaitForRecheck
+	default:
+		// coalesced: trigger already buffered; existing gate covers this wakeup
+	}
+	a.async.readyMu.Unlock()
+}
+
+// recheckWorker — single goroutine, no concurrent rechecks.
+func (a *Manager) recheckWorker() {
```
thomas-nguy: Move to another file?

```suggestion
func (worker *RecheckWorker) run() {
```

app/mempool/manager.go
```diff
@@ -243,13 +263,78 @@ func (a *Manager) StageRecheckSenders(height int64, txs [][]byte) {
...
+				a.RecheckTxs()
+				close(ready)
+			}
+		}
+	}
+}
+
+// Close stops the recheck worker. Idempotent; call before store teardown.
+func (a *Manager) Close() {
+	if a.async.stop == nil {
```
thomas-nguy: a.worker.stop()

app/mempool/manager.go
```diff
@@ -243,13 +263,78 @@ func (a *Manager) StageRecheckSenders(height int64, txs [][]byte) {
...
+		return
+	}
+	a.async.stopOnce.Do(func() { close(a.async.stop) })
+	<-a.async.done
+}
+
+// WaitForRecheck blocks until any in-progress recheck completes. ctx cancellation
+// unblocks it so a stuck worker cannot stall block production.
+func (a *Manager) WaitForRecheck(ctx context.Context) {
+	if a.async.trigger == nil {
```
thomas-nguy: a.worker.wait()

app/mempool/manager.go
```diff
@@ -243,13 +263,78 @@ func (a *Manager) StageRecheckSenders(height int64, txs [][]byte) {
 	a.stagingMu.Unlock()
 }
 
+// TriggerRecheck schedules async recheck (non-blocking, coalescing). Runs inline when no worker (test path).
+func (a *Manager) TriggerRecheck() {
+	if a.async.trigger == nil {
```
thomas-nguy: a.worker.recheck()

app/mempool/manager.go
JayT106: splitted the logic!

app/mempool/manager.go
JayT106: update comment,  it's a note for the test constructor.

app/mempool/manager.go
```diff
@@ -68,6 +72,13 @@ func NewManager(app *baseapp.BaseApp, encCache *EncoderCache, txEncoder sdk.TxEn
 	a.signer = signer
 	a.maxRecheckBatch = recheckBatchSize
 	a.ttlNumBlocks = ttlNumBlocks
+	// newManager (tests) leaves worker zero and runs RecheckTxs synchronously.
+	a.worker.trigger = make(chan chan struct{}, 1)
```
thomas-nguy: can you move all those to `worker.init()` ?

JayT106: done!

app/app.go
```diff
@@ -521,9 +527,17 @@ func New(
 				h.SetSignerExtractionAdapter(signerExtractor)
 			}
 			inner := h.PrepareProposalHandler()
-			// Stage gate-skipped senders for recheck so base-fee-stranded txs are
-			// evicted in ~1 block instead of waiting for the TTL.
+			// re-stage gate-skipped senders; evicted in ~1 block vs TTL wait.
 			app.SetPrepareProposal(func(ctx sdk.Context, req *abci.RequestPrepareProposal) (*abci.ResponsePrepareProposal, error) {
+				if mempoolManager != nil {
```
thomas-nguy: can we refactor this part to make it more readible?

Encapsulate the logic under proposal.go 

- Create a new struct CronosHandler (with mempool manager , default proposal etc)

- Define
```
func (h *CronosHandler) PrepareProposalHandler() sdk.PrepareProposalHandler {
	return func(ctx sdk.Context, req *abci.RequestPrepareProposal) (*abci.ResponsePrepareProposal, error) {}
```

- Change this line to 
```
app.SetPrepareProposal(cronosHandler.PrepareProposalHandler())
```

JayT106: Done in 15c8f092

app/proposal.go
```diff
@@ -131,6 +131,44 @@ func (ts *ExtTxSelector) gateBaseFee(goCtx context.Context) (*big.Int, string) {
 	return ts.baseFee, ts.evmDenom
 }
 
+// MempoolProposalHandler wraps PrepareProposal for mempool.type=app: waits
```
thomas-nguy: Comment here should describe the purpose of the MempoolProposalHandler in a general way , not the entire implementation logic

Just this comment is enough

```
// MempoolProposalHandler defines a custom PrepareProposal for cronos

File: app/proposal.go [hunk 16]
```diff
@@ -131,6 +131,44 @@ func (ts *ExtTxSelector) gateBaseFee(goCtx context.Context) (*big.Int, string) {
...
+}
+
+func (h *MempoolProposalHandler) SetMempoolManager(m *cronosmempool.Manager) {
+	h.mempoolManager = m
+}
+
+func (h *MempoolProposalHandler) PrepareProposalHandler() sdk.PrepareProposalHandler {
+	return func(ctx sdk.Context, req *abci.RequestPrepareProposal) (*abci.ResponsePrepareProposal, error) {
+		if h.mempoolManager != nil {
+			// ctx has no deadline from cosmos-sdk, so we set our own bound.
```
thomas-nguy: Comment is off, what we are doing here is to wait recheck  because it has been changed to async, and implementing a time out to not block proposal time in case it takes too long

app/app.go
thomas-nguy: Maybe all from here can be moved to `NewMempoolProposalHandler` 

so you can just pass the DefaultProposalHandler.  

`MempoolProposalHandler` becomes just a wrapper over `DefaultProposalHandler`

app/proposal.go
```diff
@@ -131,6 +131,44 @@ func (ts *ExtTxSelector) gateBaseFee(goCtx context.Context) (*big.Int, string) {
...
+			if h.mempoolManager.WaitForRecheckTimedOut(ctx, recheckWaitTimeout) {
+				// Recheck is still running: better to propose an empty block than pick txs from a stale pool.
+				telemetry.IncrCounter(1, "cronos", "mempool", "recheck", "proposal_timeout")
+				h.extSel.DrainGateSkipped() // keep extSel's state clean for the next round
+				return &abci.ResponsePrepareProposal{}, nil
+			}
+		}
+		resp, err := h.inner(ctx, req)
```
thomas-nguy, line 158: see my comment above.

`NewExtTxSelector` with blocklist handler logic should be moved there because its part of our `custom proposal handler`

JayT106: done by [3def010](https://github.com/crypto-org-chain/cronos/pull/2118/commits/3def010203405775ae4a4bedbb29faee18020818)

thomas-nguy (approved):

app/mempool/manager.go
```diff
@@ -394,6 +403,23 @@ func (a *Manager) selectTxs(snapshot []sdk.Tx, recheckSenders map[string]struct{
...
+// evictForRecheck evicts tx and folds its signers into recheckSenders, allocating
+// evictedSet/recheckSenders lazily so a no-eviction cycle stays alloc-free.
+func (a *Manager) evictForRecheck(tx sdk.Tx, evictedSet map[sdk.Tx]struct{}, recheckSenders map[string]struct{}) (map[sdk.Tx]struct{}, map[string]struct{}) {
+	a.evict(tx)
+	if evictedSet == nil {
+		evictedSet = make(map[sdk.Tx]struct{})
+	}
+	evictedSet[tx] = struct{}{}
+	for _, s := range a.signers(tx) {
+		if recheckSenders == nil {
```
songgaoye: nit: you could hoist the recheckSenders == nil check out of the inner loop, since it only needs to allocate once:
``` go
sigs := a.signers(tx)
if len(sigs) > 0 && recheckSenders == nil {
    recheckSenders = make(map[string]struct{})
}
for _, s := range sigs {
    recheckSenders[s] = struct{}{}
}
```

app/mempool/recheck_worker.go
```diff
@@ -0,0 +1,111 @@
...
+	trigger  chan chan struct{} // holds the pending run's ready gate
+	quit     chan struct{}
+	done     chan struct{}
+	stopOnce sync.Once
+	readyMu  sync.Mutex
+	ready    chan struct{}
+}
+
+// init allocates channels and launches the worker goroutine.
+func (w *recheckWorker) init(fn func()) {
```
songgaoye: nit: `recheckWorker` could store `fn` as a field instead of passing it in, since it always runs the same function.

songgaoye: nit: could we replace the init() method with a `newRecheckWorker()` constructor? init is a reserved-ish name in Go (package-level func init()),

app/mempool/recheck_worker.go
```diff
@@ -0,0 +1,111 @@
...
+	w.trigger = make(chan chan struct{}, 1)
+	w.quit = make(chan struct{})
+	w.done = make(chan struct{})
+	w.ready = make(chan struct{})
+	close(w.ready)
+	go w.run(fn)
```
songgaoye: nit: we could mirror the cosmos-sdk/cometbft "construct → Start → Stop" lifecycle here. Right now the goroutine is launched inside the constructor — splitting that into an explicit start() (we already have stop()) would keep the goroutine out of construction and make the lifecycle symmetric and clearer.

app/mempool/recheck_worker.go
```diff
@@ -0,0 +1,111 @@
...
+	"sync"
+)
+
+// recheckWorker is use to manage the recheck workflow.
+type recheckWorker struct {
+	trigger  chan chan struct{} // holds the pending run's ready gate
+	quit     chan struct{}
+	done     chan struct{}
+	stopOnce sync.Once
+	readyMu  sync.Mutex
```
songgaoye: nit: could we add a comment on `readyMu`? It does double duty — it guards the ready field and serializes `recheck()`'s send against `stop()`'s quit-close. That second role isn't obvious from the name, and narrowing the lock would silently reintroduce the stranded-gate race, so it's worth calling out.

app/proposal.go
```diff
@@ -131,13 +126,47 @@ func (ts *ExtTxSelector) gateBaseFee(goCtx context.Context) (*big.Int, string) {
...
 var _ baseapp.ProposalTxVerifier = &CacheProposalTxVerifier{}
 
-// CacheProposalTxVerifier replaces BaseApp.PrepareProposalVerifyTx (which runs
-// a full ante) with a cache lookup: ante already ran at admission and recheck
-// evicts stale txs, so PrepareProposal only needs canonical bytes. Cache hit ->
-// cached bytes; miss -> encode. Mirrors cosmos/evm. The skipped baseFee check for
-// idle senders is reapplied by ExtTxSelector's gate.
+// CacheProposalTxVerifier is used to cache encoded transactions to avoid cpu overhead during proposal.
 type CacheProposalTxVerifier struct {
 	*baseapp.BaseApp
```
songgaoye: Does it need include `*base.BaseApp`? It only uses `TxEncode`. right?

app/proposal.go
songgaoye, line 65: Do we need to clear `gateSkipped`?

app/mempool/manager.go
JayT106: done by https://github.com/crypto-org-chain/cronos/pull/2118/commits/d70df97fc26104c95a6df6753eaa4c12ce6958ee

app/mempool/recheck_worker.go
JayT106: do by https://github.com/crypto-org-chain/cronos/pull/2118/commits/d70df97fc26104c95a6df6753eaa4c12ce6958ee

JayT106: done by https://github.com/crypto-org-chain/cronos/pull/2118/commits/d70df97fc26104c95a6df6753eaa4c12ce6958ee

app/mempool/recheck_worker.go
JayT106: done by https://github.com/crypto-org-chain/cronos/pull/2118/commits/d70df97fc26104c95a6df6753eaa4c12ce6958ee

app/mempool/recheck_worker.go
JayT106: done by https://github.com/crypto-org-chain/cronos/pull/2118/commits/d70df97fc26104c95a6df6753eaa4c12ce6958ee

app/proposal.go
JayT106: fixed!

app/proposal.go
JayT106: fixed!

songgaoye:
Great work here! Non-blocking follow-up idea: since recheck uses the shared `checkState` and releases `a.mu` between candidates, concurrent `InsertTx/CheckTx` can interleave and make evictions timing-dependent. Node-local, so fine to defer — maybe a later PR to move recheck onto a dedicated branched context. cosmos/evm already has a working implementation we can reference (`mempool/rechecker.go`): it runs the ante on a branched `CacheMultiStore` instead of the shared `checkState`.

songgaoye (commented):
LGTM

songgaoye (approved):

PR #2144, opened 2026-07-17
feat(mempool): add hash-keyed sender cache to VerifyEthSig

JayT106:
### What

Adds a hash-keyed sender cache to `VerifyEthSig` so a tx's ecrecover result carries across the three phases it's decoded in — mempool preverify, `CheckTx`, and `FinalizeBlock`. Wires the cache into `app.go`'s admission pre-verify path and ante `HandlerOptions`, and pins `ethermint` to the fork commit that implements it.

### Issue

A tx's signature gets verified up to three times on its way through a block, once per phase, because each phase decodes the raw bytes into a fresh `*ethtypes.Transaction`. go-ethereum's own sig cache can't help here: it stores the recovered sender on that specific object, keyed by pointer identity, and the pointer changes every time the bytes are decoded again. The one thing that stays constant across phases is the signed bytes, and therefore the tx hash — not what go-ethereum's cache uses.

### Solution

- Bump `ethermint` to a fork commit ([crypto-org-chain/ethermint#1062](https://github.com/crypto-org-chain/ethermint/pull/1062)) adding `ante/cache.SenderCache`: a single LRU keyed by `(tx hash, signer sighash)` instead of object identity. Keying on the sighash too (not just the hash) means a cache hit still gets rejected if a caller passes a different signer scheme for the same tx.
- Wire `SenderCache` through `app.go`: constructed once at app init and shared by the admission pre-verify hook and the ante handler, so a cache entry populated during admission is reused by `CheckTx` and `FinalizeBlock` — no separate cache per phase.
- Size the cache off cosmos-sdk's own `mempool.max-txs`. Cache is a no-op when `mempool.max-txs` is 0 or negative.

### Test

- `app/mempool/sendercache_test.go`: proves `FinalizeBlock`'s ante pass hits the same cache instance the admission path populated, even after the decode cache has evicted the tx — the scenario go-ethereum's identity-keyed cache can't handle.
- `app/mempool/admission_test.go`: updated for the sender-cache wiring in the admission pre-verify path.

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -188,6 +188,7 @@ const (
 	FlagMempoolGossipTTL           = "cronos.mempool-gossip-ttl"
 	FlagMempoolTxsPerBlock         = "cronos.mempool-txs-per-block"
 	FlagMempoolTTLNumBlocks        = "cronos.mempool-ttl-num-blocks"
+	FlagMempoolTxSenderCacheSize   = "cronos.mempool-tx-sender-cache-size"
 )
 
 // recheckWaitTimeout bounds how long PrepareProposal waits for an in-flight async
@@ -348,6 +349,8 @@ type App struct {
 
 	mempoolManager *cronosmempool.Manager
 
+	senderCache *cache.SenderCache
+
 	// unsafe to set for validator, used for testing
 	dummyCheckTx bool
 }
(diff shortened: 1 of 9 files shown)


go.mod
```diff
@@ -396,8 +396,9 @@ replace (
 	github.com/crypto-org-chain/cronos-store/versiondb => github.com/crypto-org-chain/cronos-store/versiondb v0.0.0-20260716010842-1c77adad431c
 	// release/v1.16
 	github.com/ethereum/go-ethereum => github.com/crypto-org-chain/go-ethereum v1.10.20-0.20260521015249-663dca6c618e
+
 	// develop
-	github.com/evmos/ethermint => github.com/crypto-org-chain/ethermint v0.22.1-0.20260716064133-ceb48bc17115
+	github.com/evmos/ethermint => github.com/JayT106/ethermint v0.4.2-0.20260717193412-5fd39c161d36
```
JayT106: placeholder once upstream PR get merged.

JayT106:
@claude review

app/app.go
```diff
@@ -412,6 +421,14 @@ func New(
 		logger.Info("tx encode/decode cache enabled", "size", txCacheSize, "max-tx-bytes", maxTxBytes)
 		activeDecoder = cronosmempool.NewCachingDecoder(txDecoder, cronosmempool.NewDecodeCache(uint(txCacheSize), uint(maxTxBytes)))
 	}
+	senderCacheSize := resolveSizeOpt(appOpts, FlagMempoolTxSenderCacheSize, cmdcfg.DefaultMempoolTxSenderCacheSize)
+	var senderCache *cache.SenderCache
+	if senderCacheSize < 0 {
+		logger.Info("sender cache disabled")
+	} else {
+		logger.Info("sender cache enabled", "size", senderCacheSize)
+		senderCache = cache.NewSenderCache(senderCacheSize)
```
randy-cro: in the ethermint [PR](https://github.com/crypto-org-chain/ethermint/pull/1062), the `senderCache` is populated with `mempoolMaxTxs`, why is there a separate `senderCacheSize` flag here?

JayT106: you are correct, should populated with mempoolMaxTxs, Was thinking to populated with tx-cache-size, but it's derived value and not really correct. Fixed!

thomas-nguy (approved):

PR #2131, opened 2026-07-06
fix: bound message count and per-message gas in ReplayBlock query

thomas-nguy:
👮🏻👮🏻👮🏻 !!!! REFERENCE THE PROBLEM YOUR ARE SOLVING IN THE PR TITLE AND DESCRIBE YOUR SOLUTION HERE !!!! DO NOT FORGET !!!! 👮🏻👮🏻👮🏻

## Summary

The `ReplayBlock` gRPC query re-executes a batch of caller-supplied eth messages
without a persistent fee, making it an amplification vector on the public gRPC
endpoint: a single request could enqueue an unbounded number of messages, each
carrying an arbitrarily large gas limit, and burn unbounded node CPU.

This adds two bounds:

- **Message count** — reject requests with more than `MaxReplayBlockMsgs` (10000)
  messages. A real block never holds this many eth txs, so the legitimate
  `cronos_replayBlock` JSON-RPC path is unaffected.
- **Per-message gas** — reject any message whose gas limit exceeds the block gas
  limit (`ethermint.BlockGasLimit`), falling back to `DefaultReplayBlockGasCap`
  (25M, matching ethermint's default JSON-RPC gas cap) when the chain reports no
  block gas limit. A committed tx always fits within the block gas limit, so
  legitimate replay is unaffected.

The fallback also treats CometBFT's `max_gas = -1` ("no limit", surfaced as
`math.MaxUint64`) as "no limit", so the gas cap stays effective on chains that
disable the block gas limit.

# PR Checklist:

- [ ] Have you read the [CONTRIBUTING.md](https://github.com/crypto-org-chain/chain-main/blob/master/CONTRIBUTING.md)?
- [ ] Does your PR follow the [C4 patch requirements](https://rfc.zeromq.org/spec:42/C4/#23-patch-requirements)?
- [ ] Have you rebased your work on top of the latest master? 
- [ ] Have you checked your code compiles? (`make`)
- [ ] Have you included tests for any non-trivial functionality?
- [ ] Have you checked your code passes the unit tests? (`make test`)
- [ ] Have you checked your code formatting is correct? (`go fmt`)
- [ ] Have you checked your basic code style is fine? (`golangci-lint run`)
- [ ] If you added any dependencies, have you checked they do not contain any known vulnerabilities? (`go list -json -m all | nancy sleuth`)
- [ ] If your changes affect the client infrastructure, have you run the integration test?
[...]

PR #2146, opened 2026-07-21
fix(cronos): bound ReplayBlock message count during decode

JayT106:
### What

Caps the number of eth message elements the `ReplayBlockRequest` decoder will allocate, in `x/cronos/types/query.pb.go`'s generated `Unmarshal`, sourced from a new `MaxReplayBlockMsgs` constant in `x/cronos/types/query.go`. The keeper's existing count/gas bounds (#2131) now reference the same constant instead of duplicating it.

### Issue

The public native gRPC `ReplayBlock` endpoint takes an uncapped `repeated` list of `MsgEthereumTx`. An empty element encodes to just 2 bytes, so a request just under the 10MiB receive limit can pack over 5 million elements. Decoding that request allocates hundreds of MiB and can OOM-kill the process — and this happens during `ProtoCodec.Unmarshal`, before the keeper handler (and its #2131 length check) ever runs. Two concurrent requests reproduced a fatal Go runtime OOM in testing.

### Solution

- Added `MaxReplayBlockMsgs` in `x/cronos/types/query.go` as the single source of truth.
- Patched the generated `ReplayBlockRequest.Unmarshal` (`query.pb.go`) to reject once the element count exceeds the cap, before appending further elements — bounding decode-time allocation regardless of message-size limits.
- This edits generated code by hand; if `query.pb.go` is regenerated, the check must be re-added (flagged in a comment at both sites).
- `x/cronos/keeper/grpc_query.go` now references `types.MaxReplayBlockMsgs`.

### Test

- New `x/cronos/types/query_test.go`: crafts a raw wire payload with `MaxReplayBlockMsgs+1` empty elements and asserts `Unmarshal` rejects it without appending past the cap; also asserts a payload at exactly the cap still decodes.
- Existing `x/cronos/keeper/grpc_query_test.go` (`TestReplayBlockBounds`) still passes, updated to reference the relocated constant.

## Summary by CodeRabbit

* **Bug Fixes**
  * Replay block requests now enforce a maximum of 10,000 Ethereum messages during decoding.
  * Excessively large requests are rejected earlier, helping prevent excessive memory usage.
  * Added validation coverage for requests at and beyond the allowed message limit.

* **Documentation**
  * Updated the unreleased changelog with the ReplayBlock message-count limit fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 *Jul 16, 2026*
 
+### Bug fixes
+
+* [#2146](https://github.com/crypto-org-chain/cronos/pull/2146) fix(cronos): bound ReplayBlock message count during decode.
+
 ## v1.8.0-alpha
 
 ### Improvements
--- a/x/cronos/keeper/grpc_query.go
+++ b/x/cronos/keeper/grpc_query.go
@@ -18,14 +18,9 @@ import (
 	sdk "github.com/cosmos/cosmos-sdk/types"
 )
 
-const (
-	// MaxReplayBlockMsgs caps the eth messages one ReplayBlock query may execute.
-	MaxReplayBlockMsgs = 10000
-
-	// ReplayBlockGasCap caps per-message EVM gas in a ReplayBlock query.
-	// Since historical blocks may have used different limits, we use a fixed upper bound value.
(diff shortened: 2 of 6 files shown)


JayT106:
Reopening from fork branch JayT106:fix/replay-block-decode-oom

PR #2147, opened 2026-07-21
fix(cronos): bound ReplayBlock message count during decode

JayT106:
### What

Caps the number of eth message elements the `ReplayBlockRequest` decoder will allocate. The cap (`MaxReplayBlockMsgs`, defined once in `x/cronos/types/query.go`) is enforced inside the generated `Unmarshal` in `x/cronos/types/query.pb.go`, and `x/cronos/keeper/grpc_query.go` now references the same constant. `scripts/reapply-proto-patches.sh` re-applies the guard automatically after `make proto-gen`.

### Issue

The public native gRPC `ReplayBlock` endpoint takes an uncapped `repeated` list of `MsgEthereumTx`. An empty element encodes to just 2 bytes, so a request just under the 10MiB receive limit can pack over 5 million elements. Decoding that request allocates hundreds of MiB and can OOM-kill the process — during `ProtoCodec.Unmarshal`, before the keeper handler (and its existing #2131 length check) ever runs. Two concurrent requests reproduced a fatal Go runtime OOM in testing.

### Solution

- Added `MaxReplayBlockMsgs` in `x/cronos/types/query.go` as the single source of truth.
- Patched the generated `ReplayBlockRequest.Unmarshal` (`query.pb.go`) to reject once the element count exceeds the cap, before appending further elements — bounds decode-time allocation regardless of message-size limits. Total allocation is now double-bounded: by this count cap and by the existing 10MiB `MaxRecvMsgSize`.
- Added `scripts/reapply-proto-patches.sh`, called from `scripts/protocgen.sh` after `buf generate`: re-inserts the same guard into freshly generated `query.pb.go`.
- `x/cronos/keeper/grpc_query.go` now references `types.MaxReplayBlockMsgs`.

### Test

- `x/cronos/types/query_test.go`: crafts a raw wire payload with `MaxReplayBlockMsgs+1` empty elements and asserts `Unmarshal` rejects it without appending past the cap; also asserts a payload at exactly the cap still decodes.

## Summary by CodeRabbit

* **Bug Fixes**
  * ReplayBlock requests now enforce a maximum of 10,000 messages consistently, including during decoding.
  * Oversized requests are rejected safely with a clear invalid-argument response.

* **Tests**
  * Added coverage verifying enforcement during decoding and processing, including acceptance of requests at the maximum size.

* **Chores**
  * Proto regeneration now re-applies required security protections automatically.
  * Updated the changelog with an unreleased ReplayBlock message-count fix.


scripts/protocgen.sh
```diff
@@ -33,3 +33,7 @@ rm -rf github.com
 
 # TODO uncomment go mod tidy after upgrading to ghcr.io/cosmos/proto-builder v0.12.0
 # go mod tidy
+
+# Re-apply hand-maintained security patches that buf/protoc regeneration
+# would otherwise wipe. See scripts/reapply-proto-patches.sh for what and why.
+./scripts/reapply-proto-patches.sh
```
thomas-nguy, line 39: it seems this script wont run inside the `ghcr.io/cosmos/proto-builder:0.14.0` container and this `make proto-gen ` fails

JayT106: Fixed, removes the python3/bash dependency in the script.

JayT106:
@calude review

thomas-nguy (approved):

PR #2155, opened 2026-07-28
fix(mempool): size tx-cache-size and max-tx-bytes from mempool config directly

JayT106:
### What

Simplify how the tx encode/decode cache is sized in `app/app.go`, `app/mempool`, and `cmd/cronosd/config`:

- `tx-cache-size` now derives straight from `mempool.max-txs` — no cap, no `+mempool-txs-per-block` fallback.
- Removed `cronos.tx-cache-max-tx-bytes`; the per-entry byte cap now comes directly from `mempool.max_tx_bytes`.
- Removed `DeriveTxCacheSize`, `MaxDerivedTxCacheSize`, `DefaultTxCacheSize`, and `CronosConfig.TxCacheMaxTxBytes` — dead complexity now that the sizing logic collapsed to a couple of lines.

### Issue

The old derivation (`+mempool-txs-per-block`, capped at `2<<20`) and a separate `tx-cache-max-tx-bytes` knob were redundant: `mempool.max-txs` and `mempool.max_tx_bytes` are already the operator-set bounds on the pool. Two config knobs tracking the same thing invites drift.

### Solution

- `app/app.go`: `deriveTxCacheSize(mempoolMaxTxs)` returns `mempoolMaxTxs` when `>0`, else `-1` (cache disabled) — mirrors the existing rule for unbounded/disabled mempools.
- `maxTxBytes` falls back to `cmdcfg.DefaultTxCacheMaxTxBytes` (64 KiB) only when `mempool.max_tx_bytes` is unset.
- `app/mempool`: `defaultCacheSize` (= `cmdcfg.DefaultMempoolTxsPerBlock`) is kept only as the fallback for direct cache construction (tests, library callers) — `app.New` never hits it.
- `cmd/cronosd/config/config_test.go` deleted with `DeriveTxCacheSize`; replaced by a table test on `deriveTxCacheSize` in `app/tx_cache_size_test.go`.

### Test

- New `TestDeriveTxCacheSize` covers bounded, unbounded, and disabled `mempool.max-txs`.


JayT106:
@claude review

JayT106:
@claude review

cmd/cronosd/config/toml.go
```diff
@@ -15,12 +15,17 @@ disable-tx-replacement = {{ .Cronos.DisableTxReplacement }}
 disable-optimistic-execution = {{ .Cronos.DisableOptimisticExecution }}
 
 # Capacity of the sharded LRU tx encode/decode cache.
-# 0 = derive from mempool-txs-per-block at startup (2×, default 5800). -1 = disable entirely.
+# 0 = derive from mempool-txs-per-block and mempool.max-txs at startup: bounded
```
thomas-nguy: these comment are a bit too much, 

-1 = disabled
0 = default
or positive integer capped at 1048576

cmd/cronosd/config/toml.go
```diff
@@ -15,12 +15,17 @@ disable-tx-replacement = {{ .Cronos.DisableTxReplacement }}
...
+# recheck re-decodes the pool every commit), capped at 1048576; otherwise
+# 2x mempool-txs-per-block (default 5800), or -1 (disabled) when both are
+# unlimited.
 tx-cache-size = {{ .Cronos.TxCacheSize }}
 
 # Per-entry raw payload byte cap for the tx encode/decode cache. Txs larger than this
 # are decoded normally but not cached, bounding heap use against large txs.
-# Worst-case raw-byte footprint is roughly tx-cache-size * this value.
+# Worst-case raw-byte footprint is roughly 2 * tx-cache-size * this value (encode
+# and decode caches are each sized from tx-cache-size).
```
thomas-nguy: Could simplify these block of comments as well

cmd/cronosd/config/config.go
```diff
@@ -89,14 +93,34 @@ func DefaultCronosConfig() CronosConfig {
...
 	}
 }
 
+// DeriveTxCacheSize computes the default tx-cache-size from mempool-txs-per-block
+// and mempool.max-txs. mempoolMaxTxs: >0 bounded pool, 0 unlimited, <0 disabled.
+// txsPerBlock must be >= 0 (app.New validates this before calling).
+func DeriveTxCacheSize(txsPerBlock, mempoolMaxTxs int) int {
+	if mempoolMaxTxs > 0 {
+		sum := mempoolMaxTxs + txsPerBlock
+		if sum < mempoolMaxTxs || sum > MaxDerivedTxCacheSize { // overflow or absurd max-txs
```
thomas-nguy: how `sum < mempoolMaxTxs` if `txsPerBlock must be >= 0 (app.New validates this before calling)` per comment?

cmd/cronosd/config/toml.go
```diff
@@ -15,12 +15,17 @@ disable-tx-replacement = {{ .Cronos.DisableTxReplacement }}
...
 # Should not exceed mempool.max_tx_bytes. Default 65536 (64 KiB) covers >p99 of
 # EVM tx sizes.
 tx-cache-max-tx-bytes = {{ .Cronos.TxCacheMaxTxBytes }}
```
thomas-nguy: Why dont we use `max_tx_bytes` from config.toml instead of creating a new configuration?

cmd/cronosd/config/toml.go
thomas-nguy, line 31: Could you explain what this config is for?

thomas-nguy: can't we use the comet bft 
`mempool.size`?

JayT106: This one is to indicate what's the max transaction can be handled in a block. Not the mempool size. 
Doing 3 things:
- gossip-reap cap in NewReapTxsHandler — caps txs per ~500ms reap tick
- recheck-batch cap in NewManager — candidates re-validated per Commit
- a ref input to DeriveTxCacheSize for default tx-cache-size

cmd/cronosd/config/toml.go
JayT106: yes, we can, but the `max_tx_bytes` is 1MB for `mainnet` (same as the default cometBFT, please correct me if I am wrong). The default RAM size will be ~12GB in worst-cap, using the default mempool cache size (if every coming tx is close to 1MB). If the config set caching  2 << 20 txs in pool which this PR allowed. The worst cap will be 2TB RAM.

With 64KB, it worst-cap will be 725MB with default mempool tx cache size and then worst case is 128 GB (for caching 2 << 20 txs in pool) when all coming tx size are 64KB. So it will be an OOM attacking point I think.

cmd/cronosd/config/config.go
JayT106: fixed

cmd/cronosd/config/toml.go
JayT106: Updated!

cmd/cronosd/config/toml.go
JayT106: Updated!

JayT106:
@claude review

cmd/cronosd/config/config.go
```diff
@@ -21,7 +21,7 @@ type CronosConfig struct {
 	// Set to true to disable optimistic execution.
 	DisableOptimisticExecution bool `mapstructure:"disable-optimistic-execution"`
 	// Capacity of the sharded LRU tx encode/decode cache.
-	// 0 = derive from MempoolTxsPerBlock at startup (2×, or -1 when unlimited). -1 = disable.
+	// 0 = derive from mempool.max-txs at startup (cache off if unbounded/disabled). -1 = disable.
```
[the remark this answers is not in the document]
JayT106: no a valid issue now.

app/tx_cache_size_test.go
```diff
@@ -0,0 +1,22 @@
...
+		want          int
+	}{
+		{"bounded", 5000, 5000},
+		{"unbounded", 0, -1},
+		{"disabled", -1, -1},
+	}
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			if got := deriveTxCacheSize(tc.mempoolMaxTxs); got != tc.want {
+				t.Fatalf("deriveTxCacheSize(%d) = %d, want %d", tc.mempoolMaxTxs, got, tc.want)
```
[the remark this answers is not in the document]
JayT106: no a valid issue now.

cmd/cronosd/config/toml.go
JayT106: tx-cache-max-tx-bytes removed

app/app.go
```diff
@@ -354,6 +354,13 @@ type App struct {
 	dummyCheckTx bool
 }
 
+func deriveTxCacheSize(mempoolMaxTxs int) int {
```
thomas-nguy: This function does not look really useful tbh

thomas-nguy (approved):

app/app.go
```diff
@@ -395,22 +398,13 @@ func New(
...
 	maxTxBytes := cmdcfg.DefaultTxCacheMaxTxBytes
-	if v := appOpts.Get(FlagTxCacheMaxTxBytes); v != nil {
-		parsed, err := cast.ToIntE(v)
-		if err != nil {
-			panic(fmt.Errorf("invalid %s %q: %w", FlagTxCacheMaxTxBytes, v, err))
-		}
-		if parsed > 0 {
-			maxTxBytes = parsed
-		}
+	if mempoolMaxTxBytes := cast.ToInt(appOpts.Get(FlagMempoolMaxTxBytes)); mempoolMaxTxBytes > 0 {
```
thomas-nguy: it seems that `cmdcfg.DefaultTxCacheMaxTxBytes` is a dead code because CometBFT's default for that key is 1 MiB, and the SDK reads config.toml into the same viper that backs appOpts before merging app.toml

JayT106: fixed!

app/app.go
JayT106: removed!

PR #2156, opened 2026-07-28
feat(mempool)!: cache PendingTxs for mempool rpc calls

JayT106:
### What

`Manager.PendingTxs()` (`app/mempool/manager.go`) now serves a cached pool snapshot instead of walking the whole app mempool on every call, and the walk itself is unordered — it no longer pays for the priority mempool's tie-reordering pass. Adds `pendingTxCache` (`app/mempool/pending_cache.go`), invalidated on tx admission and at each committed block, a new `UnorderedTxs` dump on the forked cosmos-sdk's `PriorityNonceMempool` (`crypto-org-chain/cosmos-sdk`), and an on/off config toggle in `CronosConfig`: `mempool-pending-tx-cache-enabled`. `mempool-tx-ttl-enabled` (pool eviction by arrival height) ships alongside it as an unrelated toggle. A few `CronosConfig` fields were also renamed for clarity, and bool-flag parsing in `app.go` was hardened.

Closes [CC-262](https://cronoslabs.atlassian.net/browse/CC-262).

#### Breaking

Renamed TxCacheSize→MempoolTxCacheSize

### Issue

Every EVM RPC pending read goes through `PendingTxs()`: `txpool_content`, `eth_pendingTransactions`, the `eth_getTransactionByHash` pending fallback, and `resend`. Each call did a full `SelectBy` walk of the priority mempool, so RPC cost scaled with pool size × request rate. A burst of pending queries meant N independent walks of the same pool.

RPC pending-tx reads never needed priority order in the first place; only `ReapTxsHandler`'s gossip budget logic does (it prefix-scans by priority to decide what to include in a block).

So the fix gives RPC callers a way to dump the pool without paying for `reorderPriorityTies`, backed by a cache invalidated on every relevant mutation (tx admission, block completion) instead of a clock — cheaper to reason about and always correct.

### Solution

- **cosmos-sdk fork** (`crypto-org-chain/cosmos-sdk`, `types/mempool/`): added `UnorderedMempool`, an optional interface (mirrors the existing `ExtMempool`/`SelectBy` non-breaking pattern) with `UnorderedTxs(ctx) []sdk.Tx`, plus a package-level `UnorderedTxs` helper that falls back to `SelectBy` for mempools that don't implement it. `PriorityNonceMempool` implements it as a plain walk of `priorityIndex` — skips `reorderPriorityTies` and the sender-cursor tie-break iterator entirely.
- `app/mempool/helpers.go`: added `UnorderedPoolSnapshot`, a sibling to `PoolSnapshot` that calls the new `sdkmempool.UnorderedTxs`. `PoolSnapshot` itself is untouched and still backs `ReapTxsHandler`, which genuinely needs priority order.
- `Manager.PendingTxs()` now calls `UnorderedPoolSnapshot` instead of `PoolSnapshot`.
- `pendingTxCache` is a pure epoch-gated cache (`enabled bool` + `atomic.Uint64` epoch), no TTL/clock. `get()` holds its mutex across the pool walk so concurrent callers single-flight onto one walk; invalidation is a lock-free epoch bump so `admit()`/`CheckTxHandler` (tx admission) and `StageRecheckSenders` (block boundary) never block on it, and a scan racing an invalidation is never served stale (epoch is read before the walk, committed after).
- `admit()` and `CheckTxHandler` invalidate on tx admission, so a newly admitted tx shows up in the very next `PendingTxs()` call instead of waiting for a block boundary.
- Config: `mempool-pending-tx-cache-enabled` is a plain on/off bool in `CronosConfig`/app.toml (default `true`) gating whether `PendingTxs()` uses the cache at all; `false` always walks the pool directly.
- Renamed `TxCacheSize`→`MempoolTxCacheSize`. Dropped the dead `TxCacheMaxTxBytes` config knob (never read; real value comes from `mempool.max_tx_bytes`).
- `app.go`: extracted `parseBoolFlag` (previously inlined only for `mempool.recheck`) and reused it for the config's bool flags, so an invalid (non-bool) config value panics at startup instead of silently coercing to `false`.
- Telemetry: `cronos.mempool.pending.cache.hit` / `.miss`.

### Test

- cosmos-sdk fork: `TestPriorityNonceMempool_UnorderedTxsSnapshot` inserts txs across senders/priorities and asserts `UnorderedTxs` returns the same set as `Select` (set-equality, not order), and that the count matches `CountTx()`.
- `app/mempool/helpers_test.go`: `UnorderedPoolSnapshot` returns the same tx set as `PoolSnapshot`.
- `app/mempool/manager_test.go`: pending-cache hit/miss, invalidation at block boundary and on `admit()`/`CheckTxHandler` tx admission, concurrent single-flight, and invalidation racing an in-flight load is never swallowed.
- `app/mempool/recheck_test.go`: `TestRecheckTxs_InvalidatesPendingCacheAfterEviction` asserts `RecheckTxs` bumps the epoch again after evicting a timed-out tx, not just relying on `StageRecheckSenders`'s earlier bump.


JayT106:
@claude review

JayT106:
@claude review

app/mempool/manager.go
```diff
@@ -64,17 +64,21 @@ type Manager struct {
 	// recheckDisabled mirrors mempool.recheck=false: skips all rechecking,
 	// including TTL/expiry eviction
 	recheckDisabled bool
+	// pendingCache TTL-caches PendingTxs(). Only committed blocks invalidate it;
+	// admissions and recheck evictions stay hidden until the TTL lapses.
+	pendingCache pendingCache
```
thomas-nguy: please explain in comment what the cache is used for, not its implementation details

cmd/cronosd/config/toml.go
```diff
@@ -41,6 +41,14 @@ mempool-txs-per-block = {{ .Cronos.MempoolTxsPerBlock }}
...
 # sender never commits. 0 = disabled. Default 120.
 mempool-ttl-num-blocks = {{ .Cronos.MempoolTTLNumBlocks }}
+
+# Caches the PendingTxs() pool scan for this long, so a burst of RPC reads
+# (txpool_content, eth_pendingTransactions, etc.) collapses onto one walk of the
+# mempool.type=app pool instead of one walk per call. The cache is dropped at
+# every block, so staleness is bounded by this value within a block: a tx
+# submitted moments ago may not show up as pending yet, and eth_getTransactionCount
+# with "pending" may lag by those txs. 0 disables caching. Default "100ms".
+mempool-pending-cache-ttl = "{{ .Cronos.MempoolPendingCacheTTL }}"
```
thomas-nguy: could we derive from an existing config instead of adding "yet" again a new configuration

thomas-nguy:
Comments are too intrusive in this PR

cmd/cronosd/config/toml.go
JayT106: No, if we do,  we can only drive from `TimeoutCommit`I think , and it's from cometBFT's config setup, do you still want to do cross-config coupling for it? The default number is already considering our current chain setup unless we would like to change the target block time again in the future. So the op should not worry about this setup.

JayT106:
> Comments are too intrusive in this PR

trimmed!

app/mempool/manager.go
JayT106: updated!

cmd/cronosd/config/config.go
```diff
@@ -38,6 +38,9 @@ type CronosConfig struct {
 	// (by arrival height), draining proposal-skipped txs whose sender never commits.
 	// 0 disables.
 	MempoolTTLNumBlocks int `mapstructure:"mempool-ttl-num-blocks"`
```
thomas-nguy: MempoolTxTTL

thomas-nguy: TTL for transaction in the mempool

thomas-nguy: Change to boolean

cmd/cronosd/config/config.go
```diff
@@ -58,6 +61,8 @@ const (
 	// DefaultMempoolTTLNumBlocks evicts mempool.type=app txs older than this many
 	// blocks by arrival height, draining proposal-skipped txs that never commit.
 	DefaultMempoolTTLNumBlocks = 120
+
+	DefaultMempoolPendingCacheTTL = 100 * time.Millisecond
```
thomas-nguy: Change to boolean, 

Name is RPCPendingTxCache

cmd/cronosd/config/config.go
```diff
@@ -38,6 +38,9 @@ type CronosConfig struct {
 	// (by arrival height), draining proposal-skipped txs whose sender never commits.
 	// 0 disables.
 	MempoolTTLNumBlocks int `mapstructure:"mempool-ttl-num-blocks"`
+	// MempoolPendingCacheTTL caches the PendingTxs() pool-scan result for this long,
+	// so repeated RPC reads don't each re-walk the pool. 0 disables caching.
+	MempoolPendingCacheTTL time.Duration `mapstructure:"mempool-pending-cache-ttl"`
```
thomas-nguy: Change to boolean,

Name is RPCPendingTxCache

cmd/cronosd/config/config.go
thomas-nguy: MaxTxPerBlock

thomas-nguy:
```suggestion
	MempoolTxCacheSize int `mapstructure:"tx-cache-size"`
```

thomas-nguy:
```suggestion
	MempoolTxCacheMaxTxBytes int `mapstructure:"tx-cache-max-tx-bytes"`
```

app/mempool/manager.go
```diff
@@ -169,11 +172,15 @@ func (a *Manager) InsertTx(txBytes []byte) (*sdk.TxResponse, error) {
...
 }
 
+// PendingTxs returns a snapshot of pooled txs, safe for the caller to mutate.
+// It may lag the live pool by up to the configured cache TTL.
 func (a *Manager) PendingTxs() []sdk.Tx {
 	if a.mpool == nil {
 		return nil
 	}
-	return PoolSnapshot(context.Background(), a.mpool)
+	return a.pendingCache.get(func() []sdk.Tx {
```
thomas-nguy: Right now, I am not sure if the tradeoff can be justified. 

I think it would be better to update the cache on fly when new tx enter the mempool and invalidate it after the proposal phase (already done)

The reason is that checking if tx is pending could be part of the "hot path" and if the cache has a delay (ttl) this may ruin the purpose of this api. 

Setting ttl too low would make the cache not useful, and setting too high would add latency 

```
┌──────────────────┬─────────────────────────────────────────────┬──────────────────────────────────────┐
│     Consumer     │                  Endpoint                   │     100ms staleness acceptable?      │
├──────────────────┼─────────────────────────────────────────────┼──────────────────────────────────────┤
│ txpool/api.go:68 │ txpool_content                              │ Yes — this is the PR's stated target │
├──────────────────┼─────────────────────────────────────────────┼──────────────────────────────────────┤
│ call_tx.go:81    │ eth_resend                                  │ Yes, and it's rare                   │
├──────────────────┼─────────────────────────────────────────────┼──────────────────────────────────────┤
│ tx_info.go:144   │ eth_getTransactionByHash (pending fallback) │ No                                   │
├──────────────────┼─────────────────────────────────────────────┼──────────────────────────────────────┤
│ utils.go:93      │ eth_getTransactionCount(addr, "pending")    │ No                                   │
└──────────────────┴─────────────────────────────────────────────┴──────────────────────────────────────┘

```

cmd/cronosd/config/toml.go
```diff
@@ -16,31 +16,23 @@ disable-optimistic-execution = {{ .Cronos.DisableOptimisticExecution }}
...
-# are decoded normally but not cached, bounding heap use against large txs.
-# Worst-case raw-byte footprint is roughly tx-cache-size * this value.
-# Should not exceed mempool.max_tx_bytes. Default 65536 (64 KiB) covers >p99 of
-# EVM tx sizes.
-tx-cache-max-tx-bytes = {{ .Cronos.TxCacheMaxTxBytes }}
-
-# Re-gossip suppression window for mempool.type=app. A tx reaped for gossip is
-# not re-broadcast until this elapses, stopping the AppReactor from flooding the
-# whole pool to peers every reap_interval (~500ms). Default "15s".
+tx-cache-size = {{ .Cronos.MempoolTxCacheSize }}
```
thomas-nguy: should we add mempool-* to be consistent?

cmd/cronosd/config/toml.go
```diff
@@ -16,31 +16,23 @@ disable-optimistic-execution = {{ .Cronos.DisableOptimisticExecution }}
...
+
+# Per-entry raw payload byte cap for the tx encode/decode cache. Default 65536 (64 KiB).
+tx-cache-max-tx-bytes = {{ .Cronos.MempoolTxCacheMaxTxBytes }}
```
thomas-nguy: same here

cmd/cronosd/config/toml.go
JayT106: added!

cmd/cronosd/config/toml.go
JayT106: added!

randy-cro:
The tradeoff is staleness: a tx admitted moments ago can be invisible for up to the cache's TTL, so eth_getTransactionCount(addr, "pending") may lag by those txs. The fixed internal TTL is 100ms — well under block time.

^ dont think this is about block time? it is about how often the pending tx rpc is being called? if it is called very quickly, a second tx sent with the same nonce as the first one will replace it

app/mempool/manager.go
```diff
@@ -273,10 +280,11 @@ func (a *Manager) CheckTxHandler() sdk.CheckTxHandler {
 	}
 }
 
-// StageRecheckSenders records the senders of the just-committed block's txs so
-// RecheckTxs can re-validate only their remaining pending txs, and stages the
-// committed height.
+// StageRecheckSenders records committed-block senders for RecheckTxs and stages
+// the committed height.
 func (a *Manager) StageRecheckSenders(height int64, txs [][]byte) {
```
randy-cro, line 286: not too related but we probably wanna rename this function. it does 3 things now. 1. invalidate pendingCache. 2. collect senders based on committed txs. 3. stage the height

randy-cro:
image
we probably need to invalidate cache upon eviction too

randy-cro:
So from what i understand, the context here is really about trying to improve the performance of the node when there is high load on the json rpcs such as (`eth_getTransactionCount(addr, "pending")`) during high tps. As such, the bottleneck here is actually on the priorityNonceMempool i feel as the PoolSnapshot does an reordering everytime it is called, and it locks the whole mempool mutex. 

So wouldnt it be better to: 
1. Add a flag to return the mempool txs directly without priority ordering because as i know the eth_ apis dont need, only the ReapTxHandler needs when gathering txs
2. remove the ttl because at high tps, 100ms may be stale data for the rpc clients, lowering essentially means removing it
3. can still have the cache but invalidate it upon new txs entering or when the block completes


Btw can we update the PR description with this reasoning so that it is clear why are we even adding the cache in the first place

app/mempool/manager.go
JayT106: will do it in separate PR.

JayT106:
> So from what i understand, the context here is really about trying to improve the performance of the node when there is high load on the json rpcs such as (`eth_getTransactionCount(addr, "pending")`) during high tps. As such, the bottleneck here is actually on the priorityNonceMempool i feel as the PoolSnapshot does an reordering everytime it is called, and it locks the whole mempool mutex.
> 
> So wouldnt it be better to:
> 
> 1. Add a flag to return the mempool txs directly without priority ordering because as i know the eth_ apis dont need, only the ReapTxHandler needs when gathering txs
> 2. remove the ttl because at high tps, 100ms may be stale data for the rpc clients, lowering essentially means removing it
> 3. can still have the cache but invalidate it upon new txs entering or when the block completes
> 
> Btw can we update the PR description with this reasoning so that it is clear why are we even adding the cache in the first place

update PR to align with the suggestion.

JayT106:
> image we probably need to invalidate cache upon eviction too

fixed.

JayT106:
@claude review

cmd/cronosd/config/config.go
```diff
@@ -22,7 +22,7 @@ type CronosConfig struct {
 	DisableOptimisticExecution bool `mapstructure:"disable-optimistic-execution"`
 	// Capacity of the sharded LRU tx encode/decode cache.
 	// 0 = derive from mempool.max-txs at startup (cache off if unbounded/disabled). -1 = disable.
-	TxCacheSize int `mapstructure:"tx-cache-size"`
+	MempoolTxCacheSize int `mapstructure:"mempool-tx-cache-size"`
```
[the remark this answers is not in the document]
JayT106: memtioned in PR body

app/mempool/manager.go
JayT106: [cc-309](https://cronoslabs.atlassian.net/jira/software/c/projects/CC/boards/911?selectedIssue=CC-309)

app/mempool/pending_cache.go
```diff
@@ -0,0 +1,51 @@
...
+	"sync/atomic"
+
+	"github.com/cosmos/cosmos-sdk/telemetry"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+// pendingCache caches a PendingTxs() snapshot so concurrent RPC readers
+// single-flight onto one pool walk instead of one each. Invalidated purely on
+// tx admission and block completion.
+type pendingCache struct {
```
randy-cro: would `pendingTxCache` be a better name for this?

app/app.go
```diff
@@ -450,13 +464,12 @@ func New(
...
 	}
 	ttlNumBlocks := int64(cmdcfg.DefaultMempoolTTLNumBlocks)
-	if v := appOpts.Get(FlagMempoolTTLNumBlocks); v != nil {
-		// Strict parse: a silent negative is meaningless; 0 explicitly disables.
-		parsed, err := cast.ToInt64E(v)
-		if err != nil || parsed < 0 {
-			panic(fmt.Errorf("invalid %s %q: must be a non-negative integer", FlagMempoolTTLNumBlocks, v))
-		}
-		ttlNumBlocks = parsed
+	if v := appOpts.Get(FlagMempoolTxTTL); v != nil && !parseBoolFlag(FlagMempoolTxTTL, v) {
```
randy-cro: for `cronos.mempool-tx-ttl`, can we actually change it to something like `cronos.mempool-tx-ttl-enabled`, so that it is clear this is a `boolean` config and not a numerical config. there are many mempool config and for example `cronos.mempool-gossip-ttl` is numeric, so this change is meant to make it clearer for any user to set

app/app.go
```diff
@@ -450,13 +464,12 @@ func New(
...
+		ttlNumBlocks = 0
+	}
+	pendingCacheEnabled := true
+	if v := appOpts.Get(FlagRPCPendingTxCache); v != nil {
```
randy-cro: `cronos.rpc-pending-tx-cache` can consider `cronos.rpc-pending-tx-cache-enabled` too

cmd/cronosd/config/toml.go
```diff
@@ -15,9 +15,8 @@ disable-tx-replacement = {{ .Cronos.DisableTxReplacement }}
 disable-optimistic-execution = {{ .Cronos.DisableOptimisticExecution }}
 
 # Capacity of the sharded LRU tx encode/decode cache.
-# -1 = disabled, 0 = derived from mempool.max-txs (-1 if unbounded or disabled),
-# or a positive integer.
-tx-cache-size = {{ .Cronos.TxCacheSize }}
+# 0 = derive from mempool.max-txs at startup (cache off if unbounded/disabled). -1 = disable.
+mempool-tx-cache-size = {{ .Cronos.MempoolTxCacheSize }}
```
randy-cro, line 19: is there a reason why this `mempool` related configs are not under `mempool.xxx`. is it because it clashes with cometbft's namespace?

cmd/cronosd/config/toml.go
```diff
@@ -28,13 +27,17 @@ mempool-gossip-ttl = "{{ .Cronos.MempoolGossipTTL }}"
...
 # independent of per-tx TimeoutHeight (EVM txs have TimeoutHeight 0 = never
 # expire). Drains txs the proposal keeps skipping (baseFee gate, blocklist) whose
-# sender never commits. 0 = disabled. Default 120.
-mempool-ttl-num-blocks = {{ .Cronos.MempoolTTLNumBlocks }}
+# sender never commits. Default true. false disables.
+mempool-tx-ttl = {{ .Cronos.MempoolTxTTL }}
+
+# Caches the PendingTxs() pool-scan result, invalidated on tx admission and
+# block completion. Default true. false always walks the pool.
+rpc-pending-tx-cache = {{ .Cronos.RPCPendingTxCache }}
```
randy-cro: hmm i think `mempool-pending-tx-cache-enabled` would be a better name

randy-cro:
just a side comment on the mempool config, just wanna check if all them are actually needed?

randy-cro (approved):

thomas-nguy (approved):

CHANGELOG.md
```diff
@@ -8,9 +8,14 @@
 
 *Jul 16, 2026*
 
+### Improvements
+
+* [#2156](https://github.com/crypto-org-chain/cronos/pull/2156) feat(mempool): cache PendingTxs for mempool rpc calls.
```
thomas-nguy: same comment there, it should be moved to unreleased

cmd/cronosd/config/toml.go
JayT106: yes.

CHANGELOG.md
JayT106: moved

cmd/cronosd/config/toml.go
JayT106: renamed

app/app.go
JayT106: renamed

app/app.go
JayT106: renamed

app/mempool/pending_cache.go
JayT106: rename to

PR #2175, opened 2026-08-06
fix(rpc): guard `TxsResults` length before indexing by block tx position

JayT106:
### What

`x/cronos/rpc` paired two independent RPC results — `resBlock.Block.Txs` and `blockRes.TxsResults` — then indexed `TxsResults[i]` positionally in both the receipt loop and the `ReplayBlock` loop, with no length check. A short `TxsResults` panics the RPC handler with an index-out-of-range.

### Solution

- Add `checkTxsResultsLength`, called once in `getBlockDetail` — the only place the two results are paired and the sole feeder of both loops.
- Reject any length mismatch, not just a short one: CometBFT's `state/execution.go` enforces `len(block.Txs) == len(TxResults)` before commit, so an unequal pair can only come from an inconsistent RPC layer (proxy/multiplexer, or a re-sync/pruning race between the two calls).
- Error is returned, not panicked, and carries the height plus both lengths.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an RPC issue that could cause failures when transaction results did not match the number of transactions in a block.
  * RPC requests now return a clear error for mismatched transaction data instead of potentially triggering a panic.
  * Improved error handling for block detail requests when transaction data or base-fee information cannot be retrieved.

* **Tests**
  * Added coverage for matching transaction counts and both types of mismatches.


JayT106:
@claude review

x/cronos/rpc/api_test.go
```diff
@@ -0,0 +1,48 @@
...
+	t.Run("fewer tx results than txs returns error instead of panicking", func(t *testing.T) {
+		err := checkTxsResultsLength(newResBlock(3), newBlockRes(2))
+		require.Error(t, err)
+		require.Contains(t, err.Error(), "mismatched tx results length")
+	})
+
+	t.Run("more tx results than txs returns error", func(t *testing.T) {
+		err := checkTxsResultsLength(newResBlock(2), newBlockRes(3))
+		require.Error(t, err)
+	})
```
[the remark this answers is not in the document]
JayT106: fixed

x/cronos/rpc/api.go
```diff
@@ -104,7 +113,10 @@ func (api *CronosAPI) getBlockDetail(blockNrOrHash rpctypes.BlockNumberOrHash) (
 	blockHash = common.BytesToHash(resBlock.Block.Header.Hash()).Hex()
 	blockRes, err = api.backend.TendermintBlockResultByNumber(&blockNumber)
 	if err != nil {
-		api.logger.Debug("failed to retrieve block results", "height", blockNum, "error", err.Error())
+		api.logger.Debug("failed to retrieve block results", "height", blockNumber, "error", err.Error())
+		return resBlock, blockNumber, blockHash, blockRes, baseFee, err
+	}
+	if err = checkTxsResultsLength(resBlock, blockRes); err != nil {
```
thomas-nguy, line 119: You are still returning resBlock and blockRes which could be missuses,  perhaps some refactoring could be done to clarify the semantic of `resBlock` and `resBlock` ?

Do we allow them to have different length or not?

JayT106: renamed to `block` and `blockResults`, and for the returning only return the `err` info.
This is to match with the consensus raw data check, the length should always the same. if the data mismatch, meaning we need to migrate/patch the data like fix-unlucky-tx in v0.7 before.

thomas-nguy (approved):

PR #2176, opened 2026-08-06
fix(app): retry block list decryption instead of caching the blob before it is applied

JayT106:
### What

`ProposalHandler.SetBlockList` recorded the raw blob in `h.lastBlockList` *before* decrypting and parsing it. If decrypt, unmarshal, or bech32 conversion failed, the blob was already cached, so the `bytes.Equal(h.lastBlockList, blob)` fast path in the next `RefreshBlockList` short-circuited — permanently. Since `RefreshBlockList` runs from the EndBlocker every block, one failed parse left blocklist enforcement stuck at its previous state for the life of the process, with no retry and no way to recover short of a restart.

### Solution

- Record `lastBlockList` only after the blob has actually been applied: `nil` on the empty-blob branch, and after `h.blocklist = m` on the success path.
- A failure now leaves `lastBlockList` untouched, so the next block retries the same blob.
- The empty-blob case still short-circuits correctly on repeat, since `bytes.Equal(nil, []byte{})` is true — no per-block rework.

Cost while a bad blob sits on chain: one age decrypt + unmarshal per block. Cheap, and only happens in an already-misconfigured state that an operator must fix.

### Test

`TestSetBlockListRetriesAfterFailure` calls `SetBlockList` twice with the same corrupt blob, asserts both error, then asserts a valid blob succeeds. Verified it fails against the pre-fix logic (second call returned `nil` — silently "succeeding").

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 ### Bug fixes
 
+* [#2175](https://github.com/crypto-org-chain/cronos/pull/2175) fix(app): retry block list decryption instead of caching the blob before it is applied.
 * [#2155](https://github.com/crypto-org-chain/cronos/pull/2155) fix(mempool): size tx-cache-size and max-tx-bytes from mempool config directly.
 * [#2169](https://github.com/crypto-org-chain/cronos/pull/2169) fix(cronos): add safe multiply int check during voucher conversion to EVM coins.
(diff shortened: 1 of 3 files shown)


JayT106:
@claude review

thomas-nguy (approved):

PR #2177, opened 2026-08-06
fix(mempool): clear the ante nonce cache on eviction and bound it independently

JayT106:
### What

Two problems with ethermint's `AnteCache`, which tracks `(sender, nonce)` pairs so the ante handler can tell a legitimate replacement from a duplicate:

1. **Stale entries after eviction.** When the app mempool `Manager` evicts a tx — TTL sweep in `evictForRecheck`, or a `RunTx(ReCheck)` failure — it removed the tx from the mempool and the encode cache but left the `AnteCache` entry behind. The sender was then locked out of that nonce with no tx in the mempool to replace, until an unrelated LRU eviction happened to clear it. With `mempool.max-txs = 0` the cache was unbounded, so that eviction never came and the lockout was permanent.
2. **Unbounded cache.** `AnteCache` was sized from `mempoolMaxTxs`, where `0` means "unbounded mempool" — and ethermint reads `0` as "no cap on the cache", so the cache grew without limit.

### Solution

- Wire the `AnteCache` handle into `Manager` (`anteCache` field + `SetAnteCache`, mirroring the existing `SetPreVerify`). `AnteCache` is now constructed once in `app.New` before the `baseAppOptions` closure, and the same instance goes to both the Manager and `setAnteHandler` (whose signature takes the cache instead of `mempoolMaxTxs`).
- `Manager.evict` — the single function behind both eviction paths — now calls `evictAnteCache`, which walks `tx.GetMsgs()` for `*evmtypes.MsgEthereumTx` and deletes `(ethTx.GetFrom().String(), asTx.Nonce())`. Key derivation is identical to ethermint's `ante/eth.go` `CheckAndSetEthSenderNonce`, so the delete actually hits. Nil-safe and multi-message-safe.
- When `mempoolMaxTxs == 0`, bound the cache at `cmdcfg.DefaultMempoolTxsPerBlock` — already the codebase's convention for the same "size 0" case in the encoder/decoder caches. Negative still maps to the existing no-op cache.

Scope notes: a sender can only ever clear their own nonce slot, so this adds no cross-sender griefing vector. The block-inclusion removal path bypasses `Manager.evict`, but is already self-cleaning via `CheckAndSetEthSenderNonce`'s own `Delete` during DeliverTx.

### Test

`app/mempool/antecache_evict_test.go` covers `evict` directly plus both real paths (TTL sweep via `RecheckTxs`, and the recheck-failure branch), nil cache, non-EVM message, and multi-message txs. `app/antecache_bound_test.go` asserts the actual bound through the exported accessor for max-txs zero / positive / negative. Verified `TestEvict_ClearsAnteCacheEntry` fails with `evictAnteCache` disabled.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved transaction nonce-cache cleanup when mempool transactions are evicted, expire, or fail rechecking.
  - Added independent cache capacity limits, including sensible defaults when no transaction limit is configured.
  - Ensured transaction replacement settings correctly control cache behavior.
  - Improved handling of transactions containing multiple Ethereum messages and non-Ethereum transactions.

- **Tests**
  - Added coverage for cache capacity, eviction, expiration, recheck failures, replacement-disabled behavior, and safe operation without a cache.


JayT106:
@claude review

randy-cro (approved):

PR #2179, opened 2026-08-06
fix(cronos): bound concurrent `ReplayBlock` queries and honor the request context

JayT106:
### What

`/cronos.Query/ReplayBlock` is unauthenticated and runs real EVM execution — up to `ReplayBlockGasCap` (60M, doubled to a 120M gas budget) and `MaxReplayBlockMsgs` (10000) messages per call. Those per-call caps were already in place, but nothing capped how many calls ran at once, so N concurrent requests meant N × 120M gas of parallel execution. The handler also never looked at the request context, so a client that disconnected or timed out mid-batch left the node grinding through the remaining messages for nothing.

### Solution

- A package-level semaphore caps concurrent replays at 4 (`replayBlockConcurrency`). Acquire is a `select` against the caller's context, so a cancelled or expired request gives up instead of waiting.
- Bounding compute alone isn't enough: waiters were unbounded, and each holds a decoded request (cosmos-sdk's gRPC server sets no `MaxConcurrentStreams`, so an attacker can open arbitrarily many). An atomic counter now bounds running + queued at 16 and rejects beyond that with `ResourceExhausted`. The check is a single `AddInt32` compared against the bound with rollback on rejection — no check-then-act window, so concurrent callers can't overshoot.
- `ctx.Err()` is checked once per message inside the replay loop, converted via `status.FromContextError`, so a disconnect aborts the batch instead of running it out.

Tradeoff: `cronos_replayBlock` over JSON-RPC shares this pool, so heavy debugging use can now be rejected rather than queued. 4 is deliberately conservative for a heavy debug query.

### Test

`grpc_query_replay_concurrency_test.go` covers the wait path (asserts the gRPC code, not wall-clock time — the request carries zero messages, so `DeadlineExceeded` is reachable only through the semaphore `select`), the proceed path, and the queue-full rejection via test-only counter hooks in `export_test.go`. Counter balance was traced across rejection, cancellation, panic, and every early return.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved `ReplayBlock` reliability by limiting concurrent requests and rejecting excess queued requests with a clear resource-exhaustion response.
  * Requests now stop promptly when clients disconnect or cancel them.
* **Tests**
  * Added coverage for concurrency limits, queue capacity handling, request timeouts, and cancellation behavior.
* **Documentation**
  * Updated the unreleased changelog with these fixes.


JayT106:
@claude review

thomas-nguy:
@claude

thomas-nguy (approved):

PR #2186, opened 2026-08-12
fix(mempool): fall back to app.ChainID() for EVM sig preverifier when chain-id flag is unset

JayT106:
### What

`app/app.go`'s mempool setup now registers the EVM signature preverifier (`appmempool.NewEVMSigPreVerifier`) with `app.ChainID()` instead of the raw `--chain-id` flag value.

### Issue

The flag is empty on a bare `cronosd start` — the chain ID is only known from genesis at `InitChain` time. `ethermint.ParseChainID` fails on an empty string, so `NewEVMSigPreVerifier` silently returns `nil` and the preverifier never registers. Every tx admission then pays the full ecrecover cost inside the admission mutex instead of hitting the sender cache populated by the (never-running) preverifier — this was found while diagnosing a mempool-admission throughput ceiling.

`newApp()` builds `baseappOptions` via `server.DefaultBaseappOptions(appOpts)`, which already resolves the flag-or-genesis chain ID and applies it with `baseapp.SetChainID` — and that option runs before this closure, since `baseapp.NewBaseApp` applies options in slice order. So `app.ChainID()` at this point already equals.

### Solution

- Call `appmempool.NewEVMSigPreVerifier(app.ChainID(), activeDecoder, senderCache)` directly.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved EVM transaction signature verification by using the application’s chain ID when no chain ID is explicitly configured.
  - Improved transaction processing consistency by sharing ante-cache configuration across application components.
  - Added an unreleased changelog entry documenting the signature verification fix.


JayT106:
@claude review

app/app.go
```diff
@@ -367,6 +367,13 @@ func parseBoolFlag(flag string, v interface{}) bool {
 	return parsed
 }
 
+func resolvePreVerifierChainID(flagChainID, appChainID string) string {
```
randy-cro: hmm why do we need to do this when `app.ChainId()` would be the `flagChainId` if someone starts the node with `--chain-id = xxxx`

as such im thinking wouldnt 
```
preVerifiers.Register(appmempool.NewEVMSigPreVerifier(resolvePreVerifierChainID(app.ChainID()), activeDecoder, senderCache))
```
suffice?

randy-cro: from cosmos-sdk

Image

JayT106: right, updated to use `app.ChainID()`

randy-cro (approved):

